PR #1, opened 2023-01-16
ADR: initial design

bajtos:
- docs/architecture-decision-records: add template
- ADR: initial architecture for the Station Runtime

### DIFF
--- a/docs/architecture-decision-records/2023-01-initial-architecture.md
+++ b/docs/architecture-decision-records/2023-01-initial-architecture.md
@@ -236,3 +236,5 @@ Link to other ADRs, GitHub issues, documentation, etc.
 
 - [Station Runtime Research](https://www.notion.so/pl-strflt/2023-01-Station-Runtime-Research-c45c61a9397241bba98f0d67bafe4e5d)
   in Notion
+
+- [Proof of concepts for different runtimes](https://github.com/filecoin-station/runtime-poc)
--- a/docs/architecture-decision-records/2023-01-initial-architecture.md
+++ b/docs/architecture-decision-records/2023-01-initial-architecture.md
@@ -200,7 +200,7 @@ aligned with ProtocolLab's vision of driving technical breakthroughs in the
 open. The project is community-driven and contributes to other WASM initiatives.
 
 On the technical side, it is the only engine that supports the asynchronous
-programming model (AFAICT). It also has the best documentation.
+programming model (AFAICT). It also has the most extensive documentation.
 
 ## Consequences
 
--- a/docs/architecture-decision-records/2023-01-initial-architecture.md
+++ b/docs/architecture-decision-records/2023-01-initial-architecture.md
@@ -1,0 +1,238 @@
+# Initial Architecture
+
+> Status: PROPOSED
+
+<!--
+PROPOSED, ACCEPTED, REJECTED, DEPRECATED, SUPERSEDED BY {link-to-ADR}
+-->
+
+## Context
+
+<!--
+What is the issue that we're seeing that motivates this decision or change?
+-->
+
+Our grand vision is to enable Filecoin Station to run untrusted modules on
+consumer-grade computers operated by non-technical people. This is conceptually
+similar to how browsers allow websites to execute untrusted code.
+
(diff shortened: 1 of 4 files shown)


juliangruber (approved):
so far, so good

docs/architecture-decision-records/2023-01-initial-architecture.md
```diff
@@ -20,19 +20,127 @@ The modules should have access to computer resources - CPU, memory, storage, and
...
+- Filesystem and network access is sandboxed out of the box
+- The engine and platform APIs are designed for asynchrony and non-blocking I/O
+- WebAPI-compliant Fetch API implementation
+- WebAPIs can be easily called from Rust via [wasm-bindgen](https://crates.io/crates/wasm-bindgen) and [web-sys](https://crates.io/crates/web-sys) crates.
+
+Issues:
+
+- Missing documentation for embedders
+
+- Static JS imports are not sandboxed. Add this line to your TS script `import * as google from 'https://google.com/'` - it will execute even if network access is not allowed.
```
juliangruber, line 77: ouch! Can you run `deno bundle` to create a script that will have this `import` already inlined? I'm quite sure afterwards there will be no more network calls.

bajtos: Yeah, if we wanted to use Deno, then we should require modules to provide us a single file with the bundle of all source code.

Then we would have to implement parse that code to ensure it does not contain any static imports.

I think that should be easily doable 👍🏻

juliangruber: If we use deno, we could also automate the processes of using `deno build` as part of importing/updating the module

docs/architecture-decision-records/2023-01-initial-architecture.md
```diff
@@ -0,0 +1,240 @@
...
+
+- WebAssembly package manager offering a catalogue of WASM modules
+
+- Supports WASI (filesystem)
+
+- Popular in Web3 space (ChainSafe, Fluence, Hyperledger, and more)
+
+Issues:
+
+- No support for an asynchronous programming model & non-blocking I/A
```
juliangruber, line 164:
```suggestion
- No support for an asynchronous programming model & non-blocking I/O
```

docs/architecture-decision-records/2023-01-initial-architecture.md
```diff
@@ -0,0 +1,240 @@
...
+web-sys, then module authors would need to understand both Rust and JavaScript.
+
+That's why I believe it's better to choose a more bare-bone WASM engine.
+
+From the different WASM engine options out there, it seems that Wasmtime is best
+aligned with ProtocolLab's vision of driving technical breakthroughs in the
+open. The project is community-driven and contributes to other WASM initiatives.
+
+On the technical side, it is the only engine that supports the asynchronous
+programming model (AFAICT). It also has the best documentation.
```
juliangruber, line 203:
```suggestion
programming model (AFAICT). It also has the most extensive documentation.
```

I try to avoid words like "best", because they can mean different things to different people, and replace them with something more semantic. Does this work for you?

juliangruber (approved):

docs/architecture-decision-records/2023-01-initial-architecture.md
```diff
@@ -0,0 +1,240 @@
...
+- The runtime is not opinionated about networking & filesystem APIs. It's
+  possible to add WASI for filesystem access, but it's also easy to use
+  proprietary host APIs instead.
+
+- We have much more control over the I/O layer. It should be easier to implement
+  resource usage limiting.
+
+- It implements gas metering & limiting.
+
+- FMV is built on top of Wasmtime too. This can create synergies between our
```
patrickwoodhead, line 119: FVM

patrickwoodhead (approved):

docs/architecture-decision-records/2023-01-initial-architecture.md
juliangruber:
```suggestion
- FVM is built on top of Wasmtime too. This can create synergies between our
```

Issue #5, opened 2023-01-30
Hello world from JavaScript


bajtos:
The first step towards Zinnia runtime for JavaScript modules.

## Demo

Run a JavaScript module that prints "Hello world"

## Discussion points

- What API to use for logging? As a JS author, I would love `console.log` and friends.
- Can we hide Deno APIs (maybe delete them from the global object?) from the modules?
- What is the DX for "change code" - "run code"? Maybe `zinnia run my-module.js`?
- How are we going to test this functionality? It would be great to setup a CI pipeline for Zinnia.
- Bare-bone documentation for JS module authors to make it easy to add more content while adding new features.

## Tasks

- [x] `zinnia` CLI + argument parsing
- [x] runtime based on `deno_core`
- [x] hide Deno-related APIs
- [x] expose `console` APIs (see https://developer.mozilla.org/en-US/docs/Web/API/console)
- [x] tests
- [x] CI pipeline - build (reject compiler warnings!), lint, test
- [x] documentation for JS module authors

PR #14, opened 2023-02-01
cli: run JavaScript modules + provide `console` APIs

bajtos:
Start the implementation of Zinnia CLI, using `clap` for argument parsing.

Start the implementation of Zinnia Runtime, using `deno_core` as the JS/WASM engine.

Leverage useful bits from `deno_runtime`, like adding colours to terminal output and `console` Web-platform APIs.

As a general rule, I am trying to follow Deno as much as possible. Using the same conventions and coding style as they do, importing from their crates where possible and copying bits of code where it's difficult to import.

See https://github.com/filecoin-station/zinnia/issues/5


cli/main.rs
```diff
@@ -0,0 +1,67 @@
...
+#[command(author, version, about, long_about = None)]
+pub struct CliArgs {
+  #[command(subcommand)]
+  command: Commands,
+}
+
+#[derive(Subcommand)]
+pub enum Commands {
+  Run {
+    /// JavaScript file containing the Station Module to run
```
patrickwoodhead, line 24: Three /// here, should be two?

patrickwoodhead (approved):

cli/main.rs
```diff
@@ -0,0 +1,67 @@
...
+      println!("{} {file}", colors::green("EXECUTE"));
+      let main_module = deno_core::resolve_path(&file)?;
+      run_js_module(&main_module, &Default::default()).await?;
+      Ok(())
+    }
+  }
+}
+
+fn exit_with_error(error: AnyError) {
+  // Inspired by unwrap_or_exit<T> from https://github.com/denoland/deno/blob/main/cli/main.rs
```
juliangruber, line 53:
```suggestion
  // Inspired by unwrap_or_exit<T> from https://github.com/denoland/deno/blob/600fff79cdf5d52154344a0e3a8a523e1e21c3c1/cli/main.rs
```

Permalinks have higher chances or surviving

juliangruber (approved):
(with limited Rust knowledge) LGTM!

cli/main.rs
bajtos: Great question!

This is a documentation comment, see https://doc.rust-lang.org/reference/comments.html#doc-comments

cli/main.rs
bajtos: > Permalinks have higher chances or surviving

Yeah, I usually use permalinks too.

In this PR, I prefer to link to the latest version of the Deno source code to make it easier to find the latest version. That's why I am linking to the entire file, not only to the line range for this particular method.

juliangruber: Yes, but this entire file could be removed in the future! With a permalink you will always see at least what you saw when you wrote the code

bajtos: You are right. I reconsidered my original opinion and opened https://github.com/filecoin-station/zinnia/pull/16

PR #15, opened 2023-02-06
Automated tests (unit + integration)

bajtos:
- cli: test parsing of CLI args
- cli: integration tests spawning CLI executable

See #5

### DIFF
--- a/cli/args.rs
+++ b/cli/args.rs
@@ -1,0 +1,35 @@
+use clap::{command, Parser, Subcommand};
+
+#[derive(Parser, PartialEq, Debug)]
+#[command(author, version, about, long_about = None)]
+pub struct CliArgs {
+  #[command(subcommand)]
+  pub command: Commands,
+}
+
+#[derive(Subcommand, PartialEq, Debug)]
+pub enum Commands {
+  Run {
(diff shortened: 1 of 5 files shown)


cli/args.rs
```diff
@@ -0,0 +1,35 @@
...
+  pub command: Commands,
+}
+
+#[derive(Subcommand, PartialEq, Debug)]
+pub enum Commands {
+  Run {
+    /// JavaScript file containing the Station Module to run
+    file: String,
+  },
+}
```
bajtos, line 16: I extracted this code from `cli/main.rs`

cli/args.rs
```diff
@@ -0,0 +1,35 @@
...
+  fn run_js() {
+    let args = CliArgs::parse_from(["zinnia", "run", "mod.js"]);
+    assert_eq!(
+      args,
+      CliArgs {
+        command: Commands::Run {
+          file: "mod.js".to_string()
+        }
+      },
+    );
```
bajtos, line 33: Just a very basic test to verify that we can parse `run` arguments. The goal here is to ensure we have infrastructure to make it super easy to add more tests later.

juliangruber (commented):
nice!

Issue #7, opened 2023-01-30
Timers API


bajtos:
Implement a non-blocking API for waiting for a certain amount of time.

- [x] JavaScript API. Ideally, we should provide a standard set of APIs like `setTimeout` and `setInterval`. If we decide to implement something different, then let's create another GH issue - feature request to implement proper Web APIs. 
- [x] Documentation for module builders
- [ ] Automated tests

PR #34, opened 2023-02-09
feat: add `deno_web` APIs like `setTimeout`

bajtos:
Add JS APIs from `deno_web` crate.

Most notably:

- Events and `AbortController`
- `TextEncoder` and `TextDecoder` (will be needed for WASM)
- `File`, `FileReader` and `Blob`
- `MessageChannel`, `structuredClone` and friends
- `atob` and `atoa` (will be needed for WASM)
- `setTimeout`, `setInterval`, and friends
- `Performance` APIs
- Streams (`ReadableStream`, `WritableStream`, etc.)
- `URL` and friends

See #7

### DIFF
--- a/docs/building-modules.md
+++ b/docs/building-modules.md
@@ -38,12 +38,144 @@ Hello universe!
 Zinnia provides all standard JavaScript APIs, you can find the full list in
 [MDN web docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects).
 
-### console
+### Web APIs
+
+The following entities are defined in the global scope (`globalThis`).
+
+#### Console Standard
 
 Zinnia implements most of the `console` Web APIs like `console.log`. You can
 find the full list of supported methods in
 [Deno docs](https://deno.land/api@v1.30.3?s=Console) and more details about
 individual methods in
-[MDN web docs](https://developer.mozilla.org/en-US/docs/Web/API/console.)
+[MDN web docs](https://developer.mozilla.org/en-US/docs/Web/API/console)
 
-### More APIs are coming
+- [console](https://developer.mozilla.org/en-US/docs/Web/API/console)
+
+#### DOM Standard
+
+- [AbortController](https://developer.mozilla.org/en-US/docs/Web/API/AbortController)
+- [AbortSignal](https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal)
+- [CustomEvent](https://developer.mozilla.org/en-US/docs/Web/API/CustomEvent)
+- [Event](https://developer.mozilla.org/en-US/docs/Web/API/Event)
+- [EventTarget](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget)
+
+#### Encoding Standard
+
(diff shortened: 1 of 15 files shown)


bajtos:
@juliangruber This pull request is adding many new public APIs, significantly increasing the amount of stuff we will need to support in the future.

Most of these APIs are lightweight (e.g. AbortController, events, Performance, URL), so I don't anticipate any need to restrict their usage when we start limiting resource consumption.

Many other APIs will be needed in the future:
- `TextEncoder`, `TextDecoder` and `btoa` (or `atob`?) are required by wasm-bindgen
- Web Streams are needed for Fetch API. We will need some concept of streams anyways, I think it's best to build on a standard, even if it's not the greatest API.

Could you please review the list of the new APIs and let me know if there is anything you would prefer to remove from the global scope (make it private)?

Do you have any opinions on how to test this? I don't want to replicate the Deno test suite or Web API conformance tests. At the same time, I think it would be nice to have some sort of smoke tests to verify that the APIs are correctly exposed. But that's quite a lot of work, considering how many new APIs are added in this PR.

docs/building-modules.md
```diff
@@ -38,12 +38,140 @@ Hello universe!
...
+- [TextDecoder](https://developer.mozilla.org/en-US/docs/Web/API/TextDecoder)
+- [TextEncoder](https://developer.mozilla.org/en-US/docs/Web/API/TextEncoder)
+- [TextDecoderStream](https://developer.mozilla.org/en-US/docs/Web/API/TextDecoderStream)
+- [TextEncoderStream](https://developer.mozilla.org/en-US/docs/Web/API/TextEncoderStream)
+
+#### File API
+
+- [Blob](https://developer.mozilla.org/en-US/docs/Web/API/blob)
+- [File](https://developer.mozilla.org/en-US/docs/Web/API/File)
+- [FileReader](https://developer.mozilla.org/en-US/docs/Web/API/FileReader)
```
bajtos, line 74: I have mixed feelings about `Blob` and `File` related APIs. 

There is no access to the file system yet, so these APIs are only building blocks that cannot be meaningfully used (yet).

I also think the filesystem abstraction is not the right solution to allow modules to persist data. Something closer to a key/value store may be a much better fit. There are also LocalStorage and IndexedDB Web APIs (see [Deno's WebStorage API](https://deno.land/manual@v1.30.3/runtime/web_storage_api) and [MDN IndexedAPI](https://developer.mozilla.org/en-US/docs/Web/API/IndexedDB_API)) we may want to provide.

I am proposing to remove these three APIs for now. We can easily add them later, if there is a need.

WDYT?

bajtos: I removed these APIs in 8579744

juliangruber (approved):

bajtos:
Cool.

> What about just checking whether the global is defined? And then progress with regression tests where necessary. We could also add some more tests to one specific module, which "proves" that in general integrating modules from Deno works.

I will land this PR as-is and open another PR to add some smoke tests. I'll see if I can find a reasonably easy way for testo testn masse.

PR #40, opened 2023-02-13
fix: rework JS tests + fix timers impl

bajtos:
Refactor Runtime Web API tests - move all test code into external JavaScript files and rework the test runner to simply execute those files. This is the first step on a longer journey - I'd like to add more structure to these test files at some point in the future, e.g. using `Deno.test` or something similar.

After this change, the `setTimeout` test started to fail - the callback was never invoked!

As part of the change, I am also fixing that bug.


juliangruber (commented):
nice work!

Issue #43, opened 2023-02-13
ES Modules


bajtos:
At the moment, Zinnia does not support loading code from other files. We should implement support for ES Modules.

Tasks:

- [x] implementation
      → https://github.com/filecoin-station/zinnia/pull/205
      → https://github.com/filecoin-station/zinnia/pull/207
- [x] code cleanup
      → https://github.com/filecoin-station/zinnia/pull/208
- [x] create a follow-up issue for importing 3rd-party packages 
      → https://github.com/filecoin-station/zinnia/issues/209

bajtos:
This feature has two major parts:
1. Importing local modules that are part of the project
2. Consuming 3rd-party dependencies

The latter is tricky. I think it will be best to move it into a follow-up task.

See the discussion in Deno's Q1 Roadmap issue for a glimpse into the complexities involved: https://github.com/denoland/deno/issues/17475

bajtos:
Regarding 3rd-party dependencies: Deno is exploring novel approaches in this area. We may want to leverage what they arrive at.

https://deno.com/blog/package-json-support

PR #45, opened 2023-02-13
feat: assertion library + smoke tests for Web APIs

### LINKED ISSUE #7: Timers API

bajtos:
**feat: vendor `asserts` from Deno stdlib**
- Allow module authors to import from https://deno.land/std@0.177.0/testing/asserts.ts
- This is a short-term hack until we implement #43 

**test: add smoke tests for Web APIs**

This is a follow-up to improve test coverage of the APIs we are exposing from `deno_web` crate, see
- #34 
- #40

Resolves #7

### DIFF
--- a/runtime/vendored/asserts.bundle.js
+++ b/runtime/vendored/asserts.bundle.js
@@ -1,0 +1,808 @@
+// deno-fmt-ignore-file
+// deno-lint-ignore-file
+// This code was bundled using `deno bundle` and it's not recommended to edit it manually
+
+const { Deno  } = globalThis;
+const noColor = typeof Deno?.noColor === "boolean" ? Deno.noColor : true;
+let enabled = !noColor;
+function code(open, close) {
+    return {
+        open: `\x1b[${open.join(";")}m`,
+        close: `\x1b[${close}m`,
+        regexp: new RegExp(`\\x1b\\[${close}m`, "g")
+    };
+}
+function run(str, code) {
+    return enabled ? `${code.open}${str.replace(code.regexp, code.open)}${code.close}` : str;
+}
+function bold(str) {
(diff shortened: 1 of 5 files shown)


runtime/vendored/asserts.bundle.js
```diff
@@ -0,0 +1,808 @@
+// deno-fmt-ignore-file
+// deno-lint-ignore-file
+// This code was bundled using `deno bundle` and it's not recommended to edit it manually
+
+const { Deno  } = globalThis;
+const noColor = typeof Deno?.noColor === "boolean" ? Deno.noColor : true;
```
bajtos, line 6: I would love to enable coloured errors and coloured output from other Deno libs in general.

However, we decided to hide all Deno APIs.

What do you think @juliangruber @patrickwoodhead about exposing a very limited subset of Deno APIs, starting with `Deno.noColor`?

If we claim that Zinnia is mostly compatible with Deno, then I think this will be reasonable and justifiable.

juliangruber: To me colored output isn't important enough to complicate the implementation, at this stage of the project. But if it's easy to add, why not

juliangruber (approved):

runtime/vendored/asserts.bundle.js
bajtos: I created #49. You are right that this is not a priority now.

Issue #9, opened 2023-01-30
libp2p node


bajtos:
**It would be great to have the same Streams API for both HTTP and libp2p; see #25.**

- [x] Built-in `rs-libp2p` node - see https://github.com/libp2p/rust-libp2p
- [x] JavaScript API for dialling a libp2p protocol (e.g. `ping`). If this is too difficult, then we can start with a single `ping` API that's calling into Rust `ping` implementation under the hood.
- [x] JavaScript APIs
- [x] Documentation for module authors.

bajtos:
Closing as done. Some follow-up issues:
- #56 
- #57
- #62

PR #48, opened 2023-02-14
feat: built-in libp2p node

bajtos:
See #9 

---
_Quoting from the documentation:_

Zinnia comes with a built-in libp2p node based on [rust-libp2p](https://github.com/libp2p/rust-libp2p). The node is shared by all  Station Modules running on Zinnia. This way we can keep the number of open connections lower and avoid duplicate entries in routing tables.

The initial version comes with a limited subset of features. We will be adding more features based on feedback from our users (Station Module builders).

#### Networking stack

 - Transport: `tcp` using system DNS resolver
 - Multistream-select V1
 - Authentication: `noise` with `XX` handshake pattern using X25519 DH keys
 - Stream multiplexing: both `yamux` and `mplex`

#### `Zinnia.peerId`

Type: `string`

Return the peer id of Zinnia's built-in libp2p peer. The peer id is ephemeral, Zinnia generates a new peer id every time it starts.

#### `Zinnia.requestProtocol(remoteAddress, protocolName, requestPayload)`

 ```ts
 requestProtocol(
   remoteAddress: string,
   protocolName: string,
   requestPayload: Uint8Array,
 ): Promise<PeerResponse>;
 ```

Dial a remote peer identified by the `remoteAddress` and open a new substream for the protocol identified by `protocolName`. Send `requestPayload` and read the response payload.

The function returns a promise that resolves with a readable-stream-like object. Exact shape will be determined during implementation of this API. We want to be forward compatible here.

Notes:

 - The peer address must include both the network address and peer id.
 - The response size is limited to 10MB. Larger responses will be rejected with
   an error.
 - We will implement stream-based API supporting unlimited request & response
   sizes in the near future.

**Example**

```js
const response = await Zinnia.requestProtocol(
  "/dns/example.com/tcp/3030/p2p/12D3okowHR71QRJe5vrPm6zZXoH4K7z5mDsWWtxXpRIG9Dk8hqxk",
  "/ipfs/ping/1.0.0",
  new Uint8Array(32),
);
for (const chunk of response) {
  console.log(chunk);
}
```


docs/building-modules.md
```diff
@@ -123,8 +123,89 @@ individual methods in
...
+**TBD: How should the Stream API look like?**
+
+Ideally, we want to use Web Streams:
+
+- [ReadableStream](https://developer.mozilla.org/en-US/docs/Web/API/ReadableStream)
+- [WritableStream](https://developer.mozilla.org/en-US/docs/Web/API/WritableStream)
+
+```js
+// Stream implements both Readable and Writable stream APIs
+// (Is that possible with Web Streams? MDN does not describe any DupleStream.)
```
juliangruber, line 167:
```suggestion
// (Is that possible with Web Streams? MDN does not describe any DuplexStream.)
```

docs/building-modules.md
```diff
@@ -123,8 +123,89 @@ individual methods in
...
+  // peer address + peer id
+  "/dns/example.com/tcp/3030/p2p/12D3KooWRH7asdJasdr136asXdasd723MasdasdX3qiasda8sasd",
+  // protocol name
+  "/ipfs/ping/1.0.0",
+);
+```
+
+**TBD: How should the Stream API look like?**
+
+Ideally, we want to use Web Streams:
```
juliangruber, line 160: While this is ideal, how complex will it be, and how unstable is the tech? Would it be better for this stage of the project to use the simplest possible streams? Maybe iterators, event emitters, nodejs streams?

bajtos: Great comment!

I was thinking a lot about this topic over the weekend. Considering the complexity of integration with rust-libp2p, I decided to simplify the first version of libp2p API radically, remove all streaming and use buffers (`Uint8Array`) instead - both for the request and the response.

```ts
// exposed as Zinnia.requestProtocol
function requestProtocol(
  multiaddr: string,
  protoName: string,
  requestPayload: Uint8Array,
): Promise<Uint8Array>;
```

bajtos:
_Cross-posting the deleted content in case it is useful in the future._

---

#### `Zinnia.dialProtocol(multiaddr, protoName)`

Dial a remote peer identified by the `multiaddr` and open a new substream for
the protocol identified by `protoName`. The peer address must include both the
network address and peer id.

```js
const stream = Zinnia.dialProtocol(
  // peer address + peer id
  "/dns/example.com/tcp/3030/p2p/12D3KooWRH7asdJasdr136asXdasd723MasdasdX3qiasda8sasd",
  // protocol name
  "/ipfs/ping/1.0.0",
);
```

**TBD: How should the Stream API look like?**

Ideally, we want to use Web Streams:

- [ReadableStream](https://developer.mozilla.org/en-US/docs/Web/API/ReadableStream)
- [WritableStream](https://developer.mozilla.org/en-US/docs/Web/API/WritableStream)

```js
// Stream implements both Readable and Writable stream APIs
// (Is that possible with Web Streams? MDN does not describe any DupleStream.)
// The closest example is WebTransportDatagramDuplexStream
// https://developer.mozilla.org/en-US/docs/Web/API/WebTransportDatagramDuplexStream
// Which provides `readable` and `writable` properties to access the read/write stream.
const stream = await node.dialProtocol(targetPeer, "/ipfs/ping/1.0.0");
// write the request
const writableStream = stream.writable;
const writer = writableStream.getWriter();
await writer.write(Buffer.from("some string"));
await writer.close();

// read back the response in chunks
const readableStream = stream.readable;
for await (const chunk of readableStream) {
  const byteArray = chunk;
  // process the chunk
}
```

However, it would be great to be compatible with js-libp2p too.

```js
const stream = await node.dialProtocol(targetPeer, "/ipfs/ping/1.0.0");
// write the request
await stream.sink([Buffer.from("some string")]);
// read back the response in chunks
for await (const chunk of stream.source) {
  const byteArray = chunk.subarray();
  // process the chunk
}
```

In either case, we need to implement the follow Deno ops:

- **dial protocol**: this should return something like Deno duplex stream
  resource identified by resource id (RID)
- **write bytes to RID**: returns number of bytes written?
- **close the writable side of RID**
- **read bytes from RID**: must indicate EOF

docs/building-modules.md
juliangruber: yes! Maybe we could even implement ttfb measurements like this, if we add an option to drop after the first byte or so

bajtos: On the second thought, I'd prefer to return something else than `Uint8Array` to be forward compatible.

```ts
requestProtocol(
  multiaddr: string,
  protoName: string,
  requestPayload: Uint8Array,
): Promise<PeerResponse>;
```

Where `PeerResponse` can be modelled after Fetch API `Response` and provide a single method `arrayBuffer()`, or maybe implement an async iterator that will initially yield the entire response in one iteration.

bajtos: > Maybe we could even implement ttfb measurements like this, if we add an option to drop after the first byte or so

I'd like to implement the streaming of responses ASAP after we launch the first alpha. Streaming will allow you to measure TTFB without any additional Zinnia APIs.

bajtos:
The PR is ready for review 🙏🏻 

@juliangruber Please focus on especially on the JavaScript bits (the public API)

@tchardin @mxinden if you have the bandwidth to review the libp2p side, then I am happy to hear your feedback! You can find the libp2p bits in `feat-libp2p/ext/libp2p/peer.rs` and `feat-libp2p/ext/libp2p/peer/*.rs`.

/cc @patrickwoodhead in case you were interested

docs/building-modules.md
```diff
@@ -132,13 +132,76 @@ individual methods in
...
+  requestPayload: Uint8Array,
+): Promise<PeerResponse>;
+```
+
+Dial a remote peer identified by the `remoteAddress` and open a new substream
+for the protocol identified by `protocolName`. Send `requestPayload` and read
+the response payload.
+
+The function returns a promise that resolves with a readable-stream-like object.
+Exact shape will be determined during implementation of this API. We want to be
```
juliangruber: What do you mean by exact shape will be determined? Does it mean that extra keys will be added in the future? At first I read it as if this API wasn't implemented at all 🤔

docs/building-modules.md
```diff
@@ -174,3 +241,7 @@ Tracking issue: n/a
 #### Other
 
 - `XMLHttpRequest` Standard
+
+```
+
+```
```
juliangruber, line 247: WIP?

ext/libp2p/js/01_peer.js
```diff
@@ -0,0 +1,55 @@
+"use strict";
+
+((window) => {
+  const core = window.Deno.core;
+  const ops = core.ops;
+
+  window.__bootstrap.libp2p ??= {};
```
juliangruber, line 7: Is `??=` helpful here? Can this script be called multiple times?

ext/libp2p/lib.rs
```diff
@@ -0,0 +1,93 @@
...
+        .unwrap();
+      let rid = state.resource_table.add(default_node);
+      state.put::<DefaultNodeResourceId>(DefaultNodeResourceId(rid));
+      Ok(())
+    })
+    .build()
+}
+
+#[op]
+pub fn op_p2p_get_peer_id(state: &mut OpState) -> Result<String> {
```
juliangruber, line 45: is the `OpState` reference required to me mutable here? 🤔

juliangruber (commented):
Nice work! As far as I understand it, lgtm!

docs/building-modules.md
bajtos: You are right! I forgot to update this text yesterday; I did it in the morning:

https://github.com/filecoin-station/zinnia/pull/48/commits/e125eae74819a90427beadd313fb4791312c6d74

ext/libp2p/js/01_peer.js
bajtos: Good question 👍🏻 

The answer - I don't know 🤷🏻‍♂️

I am copying the style used by Deno, see e.g. here:
 
https://github.com/denoland/deno/blob/f3fb8ee18826e66d3896ea864d2bedaed6c79308/ext/fetch/26_fetch.js#L583-L585
```ts
  window.__bootstrap.fetch ??= {};
  window.__bootstrap.fetch.fetch = fetch;
  window.__bootstrap.fetch.handleWasmStreaming = handleWasmStreaming;
```

BTW this part will be reworked soon. Deno is moving away from this pre-ESM style, I am waiting for a new release of deno_core to adopt the new ESM style too.

This is how the code looks like now:

https://github.com/denoland/deno/blob/b15f9e60a040e2e450e7ca9971a5fc07dbf8b94c/ext/fetch/26_fetch.js#L585
```ts
export { fetch, handleWasmStreaming };
```

docs/building-modules.md
bajtos: I think that's VS Code "format on save" in action 🙈 Thanks for catching this!

ext/libp2p/lib.rs
bajtos: Yes, that's the way how Deno ops are designed. 

This is what happens when I change the reference to immutable:

```
error[E0277]: the trait bound `&OpState: Deserialize<'_>` is not satisfied
  --> ext/libp2p/lib.rs:44:1
   |
44 | #[op]
   | ^^^^^ the trait `Deserialize<'_>` is not implemented for `&OpState`
   |
   = help: the following other types implement trait `Deserialize<'de>`:
             &'a Path
             &'a [u8]
             &'a str
             ()
             (T0, T1)
             (T0, T1, T2)
             (T0, T1, T2, T3)
             (T0, T1, T2, T3, T4)
           and 193 others
note: required by a bound in `from_v8`
  --> /Users/bajtos/.cargo/registry/src/github.com-1ecc6299db9ec823/serde_v8-0.82.0/de.rs:48:6
   |
48 |   T: Deserialize<'de>,
   |      ^^^^^^^^^^^^^^^^ required by this bound in `from_v8`
   = note: this error originates in the attribute macro `op` (in Nightly builds, run with -Z macro-backtrace for more info)
```

The error could definitely be more helpful.

bajtos: Reported to Deno: 
- https://github.com/denoland/deno/issues/17894

docs/building-modules.md
bajtos: Fixed in 82f36a5

ext/libp2p/lib.rs
juliangruber: Makes sense, thank you!

juliangruber (approved):

.github/workflows/ci.yaml
```diff
@@ -27,6 +27,11 @@ jobs:
             os: ubuntu-latest
     runs-on: ${{ matrix.os }}
     steps:
+      - name: Install Protoc
+        uses: arduino/setup-protoc@ab6203da1c3118e4406048171b09238ad31ad73e # latest
+        with:
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
+
```
mxinden, line 34: Sorry for the external required dependency. Going away soon once https://github.com/libp2p/rust-libp2p/pull/3312 is merged and released.

docs/building-modules.md
```diff
@@ -132,13 +132,80 @@ individual methods in
...
+Station Modules running on Zinnia. This way we can keep the number of open
+connections lower and avoid duplicate entries in routing tables.
+
+The initial version comes with a limited subset of features. We will be adding
+more features based on feedback from our users (Station Module builders).
+
+#### Networking stack
+
+- Transport: `tcp` using system DNS resolver
+- Multistream-select V1
```
mxinden, line 148: Consider using `V1Lazy` to speed up connection establishment and stream negotiation (with the latter becoming 0RTT).

docs/building-modules.md
```diff
@@ -132,13 +132,80 @@ individual methods in
...
+- Authentication: `noise` with `XX` handshake pattern using X25519 DH keys
+- Stream multiplexing: both `yamux` and `mplex`
```
mxinden, line 150: In case `mplex` is not needed, I suggest not offering it at all. Yamux is the better multiplexer in all cases.

You might need it in case you want to talk to old js-libp2p implementations.

See https://github.com/libp2p/specs/pull/402 for rational.

ext/libp2p/peer/protocol.rs
```diff
@@ -0,0 +1,78 @@
...
+use libp2p::core::upgrade::{OutboundUpgrade, UpgradeInfo};
+use libp2p::futures::{future::BoxFuture, prelude::*};
+use libp2p::swarm::NegotiatedSubstream;
+use smallvec::SmallVec;
+
+use std::{fmt, io};
+
+use super::behaviour::RequestId;
+
+// FIXME: Can we use `[u8]` instead? How to avoid cloning when sending the data between threads?
```
mxinden, line 19: Where is the clone happening? You could consider adding a lifetime to `RequestProtocol`. Also you could consider using buffer pools. All that said, I wouldn't introduce such optimizations without previous benchmarks showing that this is an actual bottleneck.

mxinden (commented):
Thanks for the ping. Great to see rust-libp2p in action here.

Let me know in case I can be of any help.

Also let me know in case you need help upgrading to the next release (`v0.51.0`).

.github/workflows/ci.yaml
bajtos: Thank you for the heads up and a link to the PR; I'll keep track of progress there. 

Depending on `protoc` is not a big deal, but I am happy to remove it as soon as it's no longer needed.

docs/building-modules.md
bajtos: I noticed the `V1Lazy` option in the API docs, but I am not sure if it's safe to use.

We want to allow Zinnia modules to connect to any peer running a libp2p implementation that's considered supported - at minimum the latest version of js-libp2p, go-libp2p, and rust-libp2p. We also want Zinnia modules to be able to dial any protocol, especially custom project-specific protocols.

Do you think we are unlikely to encounter unexpected issues when using `V1Lazy`?

BTW, this is exactly the most valuable kind of feedback!

docs/building-modules.md
bajtos: Thank you for pointing this out! Somehow I thought that the JavaScript version of Yamux was not production ready yet. 🙈

What are the downsides of supporting both `yamux` and `mplex`? Is there a way how to configure libp2p to prefer `yamux` with a fallback to `mplex` where needed?

We want to allow connectivity with as many kinds of peers as reasonably possible. That's why I slightly prefer to support both multiplexing options. On the other hand, I expect that most Zinnia modules will be written for new projects, where the other side will also be created from scratch. Therefore, module builders should be able to tweak their peer config to support yamux.

ext/libp2p/peer.rs
```diff
@@ -0,0 +1,541 @@
...
+  id_keys: &identity::Keypair,
+) -> Result<transport::Boxed<(PeerId, StreamMuxerBox)>, noise::NoiseError> {
+  // Setup the transport + multiplex + auth
+  // Zinnia will hard-code this configuration initially.
+  // We need to pick reasonable defaults that will allow Zinnia nodes to interoperate with
+  // as many other libp2p nodes as possible.
+  let tcp_transport = libp2p::dns::TokioDnsConfig::system(
+    libp2p::tcp::tokio::Transport::new(libp2p::tcp::Config::new()),
+  )?
+  .upgrade(upgrade::Version::V1)
```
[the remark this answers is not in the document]
bajtos: > Given that you only have a single security protocol available, the multistream-select negotiation becomes 0RTT.

I am still relatively new to libp2p. Could you please explain a bit more about what is a security protocol? Is it `Noise:XX` in our case?

What are other security protocols commonly used? 

We will want to add support for WebTransport and QUIC once they become more widely available. Would that introduce another security protocol? (I vaguely remember WebTransport uses TLS-based certificates.) If yes, would such addition of another security protocol break `V1Lazy` negotiation?

ext/libp2p/peer/protocol.rs
bajtos: **For request payloads:**

1. The JS engine allocates `Uint8Array` on the GC-managed heap

2. Our `op` receives this value as a `ZeroCopyBuf`. Quoting from [Deno API docs](https://docs.rs/deno_core/0.52.0/deno_core/struct.ZeroCopyBuf.html):
   > A ZeroCopyBuf encapsulates a slice that's been borrowed from a JavaScript ArrayBuffer object. JavaScript objects can normally be garbage collected, but the existence of a ZeroCopyBuf inhibits this until it is dropped. It behaves much like an Arc<[u8](https://doc.rust-lang.org/nightly/std/primitive.u8.html)>.

   I think we may be able to use `&[u8]` instead of `ZeroCopyBuf` in our `op` function, but I am not sure.

   The concept is the same though - we get a reference to a slice pointing into the memory managed by the JS runtime.

   Now that I am writing this, I got an idea: maybe I can change `RequestPayload` to become an alias for `ZeroCopyBuf` and see how that works.

3. Our `op` converts that slice into `Vec<u8>` and that's where to clone/copy happens.

   https://github.com/filecoin-station/zinnia/blob/bc2712ac5606f3dc8621587c22f7ef13d2f387aa/ext/libp2p/lib.rs#L77

A somewhat-related discussion in Deno that may shed more light on this topic: https://github.com/denoland/deno/pull/4788

**For response payloads:**

1. We allocate `Vec<u8>` so we have a place to read the response payload into.
   https://github.com/filecoin-station/zinnia/blob/bc2712ac5606f3dc8621587c22f7ef13d2f387aa/ext/libp2p/peer/protocol.rs#L72-L73

2. The `op` returns this value to Deno runtime. The runtime allocates a new  `Uint8Array` on the GC-managed heap and copies the data there.

I want to rework this part very soon (see #56) to allow the JavaScript side to repeatedly read a chunk of data from the sub stream into a buffer provided by the JavaScript side (a.k.a BYOB). I expect that work to require changes to the `ResponsePayload`; thus I am not too worried about this response part.

bajtos:
> Also let me know in case you need help upgrading to the next release (`v0.51.0`).

@mxinden Thank you for the offer! I'll reach out in case of issues I am not able to resolve myself.

BTW, I was expecting Dependabot to open a pull request to upgrade libp2p to the next release. Do you happen to know why it did not and/or how can we change our config to fix that?

I suspect this is related to Dependabot's config option `versioning-strategy` ([docs](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#versioning-strategy)):
> - For apps, the version requirements are increased, for example: npm, pip and Composer.
> - For libraries, the range of versions is widened, for example: Bundler and Cargo.

In our workspace, we have both an app ([cli/Cargo.toml](https://github.com/filecoin-station/zinnia/blob/0f6804c8c2b6070b996e8f7dc622f8eebadde932/cli/Cargo.toml)) and several libraries (e.g. [ext/libp2p/Cargo.toml](https://github.com/filecoin-station/zinnia/blob/0f6804c8c2b6070b996e8f7dc622f8eebadde932/ext/libp2p/Cargo.toml) and [runtime/Cargo.toml](https://github.com/filecoin-station/zinnia/blob/0f6804c8c2b6070b996e8f7dc622f8eebadde932/runtime/Cargo.toml)).

Ideally, I want to build the CLI with the latest versions of dependencies but keep our libraries supporting a range of older versions that are compatible with the latest.

Do you have any suggestions or recommendations on specifying our dependency version ranges and configuring Dependabot to achieve that?

docs/building-modules.md
mxinden: In general `V1Lazy` is interoperable with `V1`. In addition go-libp2p also supports `V1Lazy`. I am not aware of an edge case that you could encounter.

(Unless you use multistream-select to negotiate multistream-select, which I doubt you would have the need for anywhere.)

docs/building-modules.md
mxinden: > Is there a way how to configure libp2p to prefer `yamux` with a fallback to `mplex` where needed?

Just make sure to list `yamux` before `mplex` in stream multiplexer setup.

Keeping `mplex` around for interoperability sounds reasonable.

ext/libp2p/peer.rs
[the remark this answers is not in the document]
mxinden: > > Given that you only have a single security protocol available, the multistream-select negotiation becomes 0RTT.
> 
> I am still relatively new to libp2p. Could you please explain a bit more about what is a security protocol? Is it `Noise:XX` in our case?

Correct.

> What are other security protocols commonly used?

On top of TCP there is also TLS. Though that is in alpha state on rust-libp2p today. All implementations support Noise, thus you should be good with Noise only.

> We will want to add support for WebTransport and QUIC once they become more widely available. Would that introduce another security protocol? (I vaguely remember WebTransport uses TLS-based certificates.)

All leveraging TLS, though out of the box, no need for you to configure.



> If yes, would such addition of another security protocol break `V1Lazy` negotiation?

It does not.

docs/building-modules.md
bajtos: Perfect! I opened a PR here: https://github.com/filecoin-station/zinnia/pull/87

docs/building-modules.md
bajtos: Thanks. Here is our configuration, I think we are good.

https://github.com/filecoin-station/zinnia/blob/cc19e97d103c630f19f1ba772e9dc34283259850/ext/libp2p/peer.rs#L189-L192

PR #52, opened 2023-02-21
feat: fetch api. closes #25

### LINKED ISSUE #25: HTTP client
Allow Station Modules to make HTTP requests (GET, POST, etc.) Ideally, we want to reuse Deno's implementation of the Fetch API from Web APIs.

~~Fetch API depends on Web Streams, which is tricky.~~

~~As a short-term shortcut, we can implement a subset of Fetch features ourselves, following the Fetch API spec as close as possible, with the intention to replace this limited implementation with a proper one from Deno at some point in the future. (Open a follow-up GH issue for that!)~~

~~**It would be great to have the same Streams API for both HTTP and libp2p; see #9.**~~

- [ ] JavaScript APIs
- [ ] Automated tests
- [ ] Documentation for module builders

juliangruber:
Closes #25


runtime/tests/js/webapis_tests.js
```diff
@@ -10,6 +10,38 @@ test("atob & btoa", () => {
   assertEquals(atob("c29tZSB0ZXh0"), "some text", `atob("c29tZSB0ZXh0")`);
 });
 
+await test("fetch", async () => {
+  const res = await fetch("https://google.com/");
+  assertEquals(res.status, 200);
+  const text = await res.text();
+  assert(text);
```
bajtos, line 17: It would be nice to check that `text` is a string and contains some fragment that's specific enough to give us confidence in response parsing & text decoding, but that's also generic enough to keep the test passing as the web page changes over time.

How about something like this?

```js
assertEquals(typeof text, "string");
assert(text.matches(/<body>/));
```

runtime/tests/js/webapis_tests.js
```diff
@@ -31,9 +63,9 @@ test("URL", () => {
 // A dummy wrapper to create isolated scopes for individual tests
 // We should eventually replace this with a proper test runner
 // See https://github.com/filecoin-station/zinnia/issues/30
-function test(name, fn) {
+async function test(name, fn) {
```
bajtos, line 66: ❗ 

Now that `test` is async, you should update all existing tests to `await test` too. We don't want dangling promises, do we?

runtime/js/99_main.js
```diff
@@ -46,7 +47,7 @@ delete Intl.v8BreakIterator;
   function runtimeStart(runtimeOptions) {
     core.setMacrotaskCallback(timers.handleTimerMacrotask);
     // core.setMacrotaskCallback(promiseRejectMacrotaskCallback);
-    // core.setWasmStreamingCallback(fetch.handleWasmStreaming);
+    core.setWasmStreamingCallback(fetch.handleWasmStreaming);
```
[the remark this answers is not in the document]
juliangruber: I also wasn't sure about this, will remove

runtime/tests/js/webapis_tests.js
juliangruber: I tested this, without `await` a failing test still produces a failure. Do you think we should still add the awaits?

runtime/runtime.rs
```diff
@@ -98,6 +114,10 @@ pub async fn run_js_module(
           "js/98_global_scope.js",
           "js/99_main.js",
         ))
+        .state(move |state| {
+          state.put(ZinniaPermissions {});
```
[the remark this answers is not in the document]
juliangruber: Before this, I got an error like this:

```
thread 'fetch_tests' panicked at 'required type ZinniaPermissions is not present in GothamState container', /Users/julian/.cargo/registry/src/github.com-1ecc6299db9ec823/deno_core-0.171.0/gotham_state.rs:82:3
```

I found this to match, so just copied the strategy: https://github.com/denoland/deno/blob/34bfa2cb2c1f0f74a94ced8fc164e81cc91cb9f4/runtime/web_worker.rs#L377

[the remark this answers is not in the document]
juliangruber: See also https://docs.rs/gotham/latest/gotham/state/index.html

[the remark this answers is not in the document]
bajtos: Makes sense. Thank you for the pointers.

I was surprised because we already use `ZinniaPermissions` for `deno_web`, but we didn't need to store it in the Gotham state. 🤷🏻

It's great to see that the tests are already helpful, even in the current minimal form.

runtime/tests/js/webapis_tests.js
bajtos: Interesting. IIUC, when a dangling promise is rejected, we enter the "unhandled rejection" path and eventually fail the test. I guess that's fine.

But why are you awaiting some tests and don't await other tests?

Is it possible to be more consistent and use the same approach everywhere?

bajtos (approved):
LGTM 👍🏻 

It would be nice to improve the JS tests to consistently use or not use `await`, see https://github.com/filecoin-station/zinnia/pull/52#discussion_r1113943644

The build was failing on `rustfmt` checks, see https://github.com/filecoin-station/zinnia/actions/runs/4235155865/jobs/7358672932 

I fixed the problem in aa0ff87. Please remember to `git pull` in your feature branch before making more changes.

runtime/tests/js/webapis_tests.js
```diff
@@ -1,39 +1,72 @@
...
+  const request = new Request("https://example.com/");
+  await request.arrayBuffer();
+});
+
+await test("Response", async () => {
+  const response = new Response();
+  await response.arrayBuffer();
+});
+
+await test("TextEncoder", () => {
```
bajtos, line 46: Oh. Would you mind moving fetch tests into a new file ( `runtime/tests/js/fetch_tests.js`)?

I'd like to start structuring our tests early and keep our test files relatively small.

As a nice side effect, you won't need to change existing tests in this file; they can stay synchronous. You will introduce async tests only in your new test file.

bajtos: Done in 9e43a95

bajtos (approved):
:shipit: 

I have enabled auto-merge, the PR should land after all CI checks pass.

Issue #53, opened 2023-02-21
Richer README


juliangruber:
- [x] It would be nice if it linked to some Notion pages, for extra context https://github.com/filecoin-station/zinnia/pull/55

The goals of the README should be (imo):
1. [x] Elevator pitch of the project https://github.com/filecoin-station/zinnia/pull/59
2. [x] How to run a rudimentary example https://github.com/filecoin-station/zinnia/pull/54
3. [x] How to create a basic module https://github.com/filecoin-station/zinnia/pull/54
4. [x] How to contribute (either through code, or telling us use cases) https://github.com/filecoin-station/zinnia/pull/55
5. [x] High level project roadmap https://github.com/filecoin-station/zinnia/pull/55

juliangruber:
Via @bajtos:

- README in workspace root - this is the landing page for our GitHub repo. https://github.com/filecoin-station/zinnia
- README for Zinnia CLI crate `zinnia` - this is the landing page on [crates.io](http://crates.io/). https://crates.io/crates/zinnia (sourced from https://github.com/filecoin-station/zinnia/blob/main/cli/README.md)

juliangruber:
> How to create a basic module

@bajtos are we ready to talk about zinnia "modules" yet? Is a module simply a js file with 0-n files it imports from?

bajtos:
> > How to create a basic module
> 
> @bajtos are we ready to talk about zinnia "modules" yet? Is a module simply a js file with 0-n files it imports from?

Yes, I think we are ready.

We don't support module imports yet (see #43), so right now a module is simply a single JS file that implements a long-running process.

I know the missing support for ES Module imports is a poor DX; I'd like to fix it soon after the first public alpha. I propose mentioning #43 in the text explaining that a module must be a single file.

Issue #22, opened 2023-02-07
Run Prettier checks as part of CI


bajtos:
Check the formatting of all JavaScript & Markdown files.

Optionally consider moving from Prettier to Deno lint, if we can find a way how to make it play nice with VS Code.

PR #55, opened 2023-02-21
docs: add contributing. #53

### LINKED ISSUE #28: link to roadmap from readme
It would be useful if the readme or another public place linked to the roadmap

juliangruber:
Closes #28

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,9 +43,16 @@ See [CLI](./cli/README.md).
 
 ## Contributing
 
-The [roadmap](https://www.notion.so/pl-strflt/Zinnia-Roadmap-b849d322502f4c079525cd8cc887df02) and [issue tracker](https://github.com/filecoin-station/zinnia/issues) are great places to check out for your first
-time to contribute! You can also look up some past decisions in [docs/architecture-decisions-records/](docs/architecture-decision-records/).
+The
+[roadmap](https://www.notion.so/pl-strflt/Zinnia-Roadmap-b849d322502f4c079525cd8cc887df02)
+and [issue tracker](https://github.com/filecoin-station/zinnia/issues) are great
+places to check out for your first time to contribute! You can also look up some
+past decisions in
+[docs/architecture-decisions-records/](docs/architecture-decision-records/).
 
-If you see something broken or missing, please [create a new issue](https://github.com/filecoin-station/zinnia/issues/new).
+If you see something broken or missing, please
+[create a new issue](https://github.com/filecoin-station/zinnia/issues/new).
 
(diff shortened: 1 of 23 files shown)


bajtos (commented):
@juliangruber could you please format the markdown using `prettier`?

We don't have a CI check to run prettier yet, but the configuration files are already in place - see e.g. [.prettierrc](https://github.com/filecoin-station/zinnia/blob/main/.prettierrc.yaml)

README.md
```diff
@@ -40,3 +40,12 @@ computers.
 ## Basic use
 
 See [CLI](./cli/README.md).
+
+## Contributing
+
+The [roadmap](https://www.notion.so/pl-strflt/Zinnia-Roadmap-b849d322502f4c079525cd8cc887df02) and [issue tracker](https://github.com/filecoin-station/zinnia/issues) are great places to check out for your first
```
bajtos, line 46: I think this is effectively closing the following issue:
- https://github.com/filecoin-station/zinnia/issues/28

bajtos:
> We don't have a CI check to run prettier yet, but the configuration files are already in place - see e.g. [.prettierrc](https://github.com/filecoin-station/zinnia/blob/main/.prettierrc.yaml)

See
- #22

README.md
juliangruber: good catch!

juliangruber:
I ran prettier, was it right like this?

runtime/vendored/asserts.bundle.js
```diff
@@ -2,786 +2,912 @@
 // deno-lint-ignore-file
 // This code was bundled using `deno bundle` and it's not recommended to edit it manually
 
-const { Deno  } = globalThis;
+const { Deno } = globalThis;
```
bajtos, line 5: We don't want to (re)format vendored files. Could you please update [.prettierignore](https://github.com/filecoin-station/zinnia/blob/main/.prettierignore) to skip everything in `runtime/vendored`?

Either in this PR, or else:
1. Revert changes in this file, land the rest of this PR.
2. Open a new PR to fix prettier ignore config.

bajtos: > I ran prettier, was it right like this?
> `npx prettier -w .`

Yes, that looks good to me!

bajtos (changes requested):
> I ran prettier, was it right like this?
> `npx prettier -w .`

Yes, that looks good to me!

Except we don't want to (re)format vendored files, see my comment below.

bajtos (approved):

PR #61, opened 2023-02-23
ci: test all crates together

bajtos:
When testing crates independently, we are not able to build `zinnia_libp2p`
(see #48):

```
   Compiling v8 v0.60.1
error: could not find native static library `rusty_v8`, perhaps an -L flag is missing?
error: could not compile `v8` due to previous error
```

In this change, I am reworking the test matrix to test all crates together. Cargo is already doing a great job in running tests in parallel.

It takes longer to build our tests than to execute them, so this change may speed up our CI builds too.


juliangruber (commented):
lgtm!

Issue #73, opened 2023-02-28
Maintenance - March'23 edition


bajtos:
- [ ] #68 

- [ ] Upgrade Deno core & runtime to the new version released on Feb 25. They have reworked internals to use ESM, which is great for us too, but requires extra work on our side.

- [ ] Upgrade rust-libp2p to v0.51.0 released on Feb 24th.

- [ ] Move `PeerNode` instance outside of Station runtime struct, so that we can share a single node among multiple runtime instances

- [ ] Make error handling more robust
   https://github.com/filecoin-station/zinnia/blob/bc2712ac5606f3dc8621587c22f7ef13d2f387aa/ext/libp2p/lib.rs#L34-L35
   https://github.com/filecoin-station/zinnia/blob/bc2712ac5606f3dc8621587c22f7ef13d2f387aa/ext/libp2p/lib.rs#L80

- [ ] Fix the libp2p test to listen on an ephemeral port assigned by the OS
   https://github.com/filecoin-station/zinnia/blob/bc2712ac5606f3dc8621587c22f7ef13d2f387aa/ext/libp2p/peer.rs#L435-L436

- [ ] Fix the JS libp2p test to connect to a local libp2p peer, not the one running on Fly.io
   https://github.com/filecoin-station/zinnia/blob/bc2712ac5606f3dc8621587c22f7ef13d2f387aa/runtime/tests/js/libp2p_tests.js#L83-L84

- [ ] Implement proper shutdown of PeerNode

   https://github.com/filecoin-station/zinnia/blob/bc2712ac5606f3dc8621587c22f7ef13d2f387aa/ext/libp2p/peer.rs#L171-L172

- [ ] #72

PR #85, opened 2023-03-02
feat: fix compat with Rust, enable Ping + Identify

bajtos:
Enable handling of `Ping` and `Identify` requests by all Zinnia nodes.

This fixes compatibility with the rust-libp2p implementation of Ping protocol, which we need to allow us to test our request-response-like behavior against a Ping protocol implementation.

This is a follow-up for
- #83 
- #84

and a part of
- #73

### DIFF
--- a/ext/libp2p/peer.rs
+++ b/ext/libp2p/peer.rs
@@ -88,7 +88,7 @@ impl PeerNode {
         // higher layer network behaviour logic.
         let swarm = Swarm::with_tokio_executor(
             tcp_transport,
-            ComposedBehaviour {
+            NodeBehaviour {
                 zinnia: RequestResponse::new(config),
             },
             peer_id,
@@ -196,7 +196,7 @@ pub fn create_transport(
 }
 
 pub struct EventLoop {
-    swarm: Swarm<ComposedBehaviour>,
+    swarm: Swarm<NodeBehaviour>,
     command_receiver: mpsc::Receiver<Command>,
     pending_dial: HashMap<PeerId, oneshot::Sender<Result<(), Box<dyn Error + Send>>>>,
     pending_requests: HashMap<RequestId, PendingRequest>,
@@ -207,7 +207,7 @@ pub struct PendingRequest {
 }
 
 impl EventLoop {
-    fn new(swarm: Swarm<ComposedBehaviour>, command_receiver: mpsc::Receiver<Command>) -> Self {
+    fn new(swarm: Swarm<NodeBehaviour>, command_receiver: mpsc::Receiver<Command>) -> Self {
         Self {
             swarm,
             command_receiver,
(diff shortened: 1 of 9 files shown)


ext/libp2p/peer.rs
```diff
@@ -422,56 +468,62 @@ mod tests {
...
-            server_transport,
-            libp2p::ping::Behaviour::new(
-                libp2p::ping::Config::new()
-                    .with_max_failures(std::num::NonZeroU32::new(10).unwrap()),
-            ),
-            server_peer_id,
-        );
-        let server_task = {
+        let listener_addr: Multiaddr = "/ip4/127.0.0.1/tcp/10458".parse().unwrap();
+        listener_swarm.listen_on(listener_addr.clone()).unwrap();
```
bajtos, line 493: @mxinden What is the canonical way for setting up a Swarm listening on an ephemeral port?

I want to write something like this - notice the port is 0 (`/tcp/0`):

```rs
let listener_addr: Multiaddr = "/ip4/127.0.0.1/tcp/10458".parse().unwrap();
listener_swarm.listen_on(listener_addr.clone()).unwrap();
// 1. wait until OS binds the port a specific number
// 2. get the updated multi address
// E.g. the function implementing (1) can return a future of the new multiaddr
```

mxinden: You can call `listen_on("/ip4/127.0.0.1/tcp/0".parse().unwrap()).unwrap()` and then wait for a `SwarmEvent::NewListenAddr`.

ext/libp2p/peer.rs
```diff
@@ -422,56 +468,62 @@ mod tests {
         init();
         let cancellation_token = CancellationToken::new();
 
-        let server_id_keys = identity::Keypair::generate_ed25519();
-        let server_peer_id = server_id_keys.public().to_peer_id();
-        let mut server_transport = create_transport(&server_id_keys).unwrap();
+        let listener_id_keys = identity::Keypair::generate_ed25519();
```
mxinden, line 471: Note that the new `libp2p-swarm-test` crate can help with all this boilerplate:

https://github.com/libp2p/rust-libp2p/tree/master/swarm-test

mxinden (commented):
Sorry for the delayed review.

ext/libp2p/peer.rs
bajtos: @mxinden thank you for the tip, the helpers in the swarm-test crate look really helpful! I would love to use [`listen`](https://github.com/libp2p/rust-libp2p/blob/32598d89d6a3008e416679ca4c2705d8b439c541/swarm-test/src/lib.rs#L77), see https://github.com/filecoin-station/zinnia/pull/85#discussion_r1123115694

There is a catch though: the swarm-test crate is not re-exported by libp2p and it's not published as a standalone crate to crates.io either :(

bajtos: Do you know how I can use `swarm-test` from my code?

Issue #99, opened 2023-03-09
Support Windows


bajtos:
- Release builds are failing on Windows, see https://github.com/filecoin-station/zinnia/issues/23#issuecomment-1456262838
- After upgrading to `deno_core@0.173` and `deno_runtime@0.99`, we don't have enough disk space in GHA runners to build Zinnia, see e.g. https://github.com/filecoin-station/zinnia/actions/runs/4374820007/jobs/7654726747

**Tasks**

- [x] Fix dev builds on Windows to make our CI pass - #106
- [x] Make the check `Build on x86_64-pc-windows-msvc` required for pull requests
- [x] Fix release builds on Windows too - #107
- [x] Run our tests on Windows for all CI runs - #112
- [x] Make the check `Test all crates (x86_64-pc-windows-msvc, windows-latest)` required for all pull requests

PR #106, opened 2023-03-14
ci: run on Windows

bajtos:
See #99


bajtos:
Nice! Looks like my expectation was correct and #105 fixed the out-of-memory problem we were experiencing on Windows GHA runners. We are one step closer to pre-built Windows binaries 💪🏻

Issue #75, opened 2023-02-28
Integration with Filecoin Station


bajtos:
StarMaps entry: https://github.com/filecoin-station/roadmap/issues/8

Once the first module is built on Zinnia, we should quickly integrate Zinnia into the Station. 

For example, as a short-term solution, we can use the current approach where each module is a standalone executable binary, with the slight modification that new modules will contain two pieces - Zinnia CLI and the module JS bundle.

On the Zinnia side, we may need to implement platform APIs for the following features:

- Get the FIL wallet address for receiving rewards.
- Publish Activity Log items.
- Report completion of jobs.

**Tasks**

- [x] Architecture decision record: 
   #121
- [x] Zinnia API for the wallet address, including docs for module builders: 
   #123 
- [x] Zinnia API for publishing activity logs, including docs for module builders.
   #124
- [x] Explain when to use `Zinnia.log.info` and when to use Console APIs like `console.log` 
   #124
  - see https://github.com/filecoin-station/zinnia/pull/121#discussion_r1143289663
- [x] Zinnia API for reporting job completion, including docs for module builders: 
   #124
- [x] Binary to be used by Station
  #141 
- [x] Configuration via env vars (wallet address, state root dir)
  #141
- [x] CD pipeline to build `zinniad` release artefacts
  #158 
  #159
- [x] A new Zinnia module that will be run by `zinniad` & the Station
  https://github.com/filecoin-station/mod-peer-checker/pull/1
- [x] Integration into Station Core, including automation to update Zinnia & modules.
  https://github.com/filecoin-station/core/pull/92
  https://github.com/filecoin-station/zinnia/pull/172
- ~~Integration into Station Desktop, including automation to update Zinnia & modules.~~ 
  → moved to https://github.com/filecoin-station/roadmap/issues/15
- ~~Documentation for module builders explaining how to add their new module to Station~~
  → we are not ready for adding 3rd-party Zinnia modules yet
- [ ] Documentation for `zinniad` (add a README)
  → https://github.com/filecoin-station/zinnia/pull/203

PR #123, opened 2023-03-21
feat: add `Zinnia.walletAddress` API

bajtos:
Allow modules to access the wallet address where to send rewards.  The value is hard-coded to a testnet address for now.

When running inside the Station, this API will return the Station wallet address.

In the future, we can allow CLI users to configure this address too.

Connect to #75

### DIFF
--- a/runtime/runtime.rs
+++ b/runtime/runtime.rs
@@ -34,6 +34,9 @@ pub struct BootstrapOptions {
 
     /// Seed value for initializing the random number generator
     pub rng_seed: Option<u64>,
+
+    /// Filecoin wallet address - typically the built-in wallet in Filecoin Station
+    pub wallet_address: String,
 }
 
 impl Default for BootstrapOptions {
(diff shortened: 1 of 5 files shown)


runtime/runtime.rs
```diff
@@ -43,6 +46,8 @@ impl Default for BootstrapOptions {
             is_tty: colors::is_tty(),
             agent_version: format!("zinnia_runtime/{}", env!("CARGO_PKG_VERSION")),
             rng_seed: None,
+            // See https://lotus.filecoin.io/lotus/manage/manage-fil/#public-key-address
+            wallet_address: "t1abjxfbp274xpdqcpuaykwkfb43omjotacm2p3za".to_owned(),
```
juliangruber, line 50:
```suggestion
            wallet_address: String::from("t1abjxfbp274xpdqcpuaykwkfb43omjotacm2p3za"),
```

The string type takes ownership, I think it's cleaner to construct it directly

juliangruber: I hope this compiles xD

bajtos: FWIW, `.to_owned()` was suggested by Rust Analyzer plugin for VS Code.

I am fine either way and will apply your suggestion.

juliangruber (approved):
Awesome!

Again I might be totally wrong, but I assume `String::from("hardcoded")` is more idiomatic than `"hardcoded".to_owned()`

PR #157, opened 2023-03-29
ci: run tests on macOS

bajtos:
We have platform-specific code in `zinniad` that should be tested on macOS too.

See #75

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -60,11 +60,8 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          # Should we run the tests on macOS too?
-          # - target: "x86_64-apple-darwin"
-          #   os: macos-latest
-          # - target: aarch64-apple-darwin
-          #   os: macos-latest
+          - target: "x86_64-apple-darwin"
+            os: macos-latest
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yaml
```diff
@@ -60,11 +60,8 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          # Should we run the tests on macOS too?
-          # - target: "x86_64-apple-darwin"
-          #   os: macos-latest
-          # - target: aarch64-apple-darwin
-          #   os: macos-latest
+          - target: "x86_64-apple-darwin"
+            os: macos-latest
```
juliangruber, line 64: while we're at it, what about adding arm64 too?

bajtos: Great call! I have considered this too.

Unfortunately, GitHub Actions don't offer GitHub-hosted macOS runners on Apple Silicon yet. This feature is currently planned for Q4 2023 - see https://github.com/github/roadmap/issues/528

I don't think we can run arm64 binaries on x64 machines, can we?

juliangruber: Right, makes sense! Let's wait :)

PR #180, opened 2023-04-13
chore: disable Dependabot for Deno crates

bajtos:
More often than not, these crates must be updated together, and Dependabot cannot do that yet (see https://github.com/github/roadmap/issues/148)


bajtos:
@juliangruber FYI

PR #195, opened 2023-04-25
docs: update the taglines

bajtos:
Update the taglines to describe Zinnia as _A sandboxed and resource-limited runtime for distributed workers._

### DIFF
--- a/runtime/lib.rs
+++ b/runtime/lib.rs
@@ -3,6 +3,8 @@ pub use deno_core;
 pub mod runtime;
 pub use runtime::*;
 
+mod module_loader;
+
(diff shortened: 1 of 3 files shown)


bajtos:
Finding a great tagline is tricky and I don't want to wait until we find the perfect one. Let's move forward with an incremental improvement:

> A sandboxed and resource-limited runtime for distributed workers.

I am happy to rework the tagline again in the future, if/when we agree on a better one.

@juliangruber PTAL

juliangruber (approved):
Agreed on the approach and this iteration 👍

Issue #144, opened 2023-03-27
zinniad: run multiple modules


bajtos:
ATM, `zinniad` can run a single module only. Before we can add another Zinnia-based module to Filecoin Station, we need to add support for running multiple modules simultaneously.

Issue #143, opened 2023-03-27
Graceful shutdown


bajtos:
We should implement a graceful shutdown in `zinniad`, preferably in `zinnia` too. The shutdown should flush out uncommitted state data and cleanly close state files.

Ideally, this shutdown should be able to interrupt JavaScript code that is currently running inside V8. (E.g. heavy computations or infinite loops.)

Note: to use this feature in Station, we need to change the way how Station Core terminates the Zinniad process.

Quoting from https://nodejs.org/api/process.html#process_signal_events:
> Windows does not support signals so has no equivalent to termination by signal, but Node.js offers some emulation with process.kill(), and subprocess.kill():
> - Sending SIGINT, SIGTERM, and SIGKILL will cause the unconditional termination of the target process, and afterwards, subprocess will report that the process was terminated by signal.
> - Sending signal 0 can be used as a platform independent way to test for the existence of a process.

bajtos:
Note: to use this feature in Station, we need to change the way how Station Core terminates the Zinniad process.

Quoting from https://nodejs.org/api/process.html#process_signal_events:
> Windows does not support signals so has no equivalent to termination by signal, but Node.js offers some emulation with process.kill(), and subprocess.kill():
> - Sending SIGINT, SIGTERM, and SIGKILL will cause the unconditional termination of the target process, and afterwards, subprocess will report that the process was terminated by signal.
> - Sending signal 0 can be used as a platform independent way to test for the existence of a process.

PR #198, opened 2023-05-02
feat: persist job counter across restarts

### LINKED ISSUE #148: zinniad: persist job counter across restarts
We should maintain a state file and keep the job counters there so that we continue after restarting where we ended before shutting down.

Related: 
- #143

bajtos:
- zinniad: persist "jobs completed" across restarts
- refactor: code cleanup in reporters

Resolves #148

The implementation is a bit hacky. We will need to refactor it to enable support for more than one module (see #144). Right now, I am optimising to deliver Zinnia in Station ASAP.

It's probably best to review the changes commit by commit.

### DIFF
--- a/daemon/main.rs
+++ b/daemon/main.rs
@@ -1,6 +1,8 @@
 mod args;
+mod state;
 mod station_reporter;
 
+use std::path::PathBuf;
 use std::rc::Rc;
 use std::time::Duration;
 
@@ -36,6 +38,9 @@ async fn run(config: CliArgs) -> Result<()> {
         ));
     }
 
+    let state_file = PathBuf::from(config.state_root).join("state.json");
+    log::debug!("Using state file: {}", state_file.display());
+
     log_info_activity("Module Runtime started.");
 
     let file = &config.files[0];
@@ -56,6 +61,7 @@ async fn run(config: CliArgs) -> Result<()> {
         ),
         wallet_address: config.wallet_address,
         reporter: Rc::new(StationReporter::new(
+            state_file,
             Duration::from_millis(200),
             module_name.into(),
         )),
--- a/daemon/state.rs
+++ b/daemon/state.rs
@@ -1,0 +1,99 @@
+use atomicwrites::{AtomicFile, OverwriteBehavior};
+use serde::{Deserialize, Serialize};
+use std::io::Write;
+use std::path::Path;
+
+#[derive(Serialize, Deserialize, Debug)]
+pub struct State {
+    pub total_jobs_completed: u64,
+}
+
+impl Default for State {
+    fn default() -> Self {
+        Self {
+            total_jobs_completed: 0,
+        }
+    }
+}
+
+pub fn load(state_file: &Path) -> State {
+    log::debug!("Loading initial state from {}", state_file.display());
+    match std::fs::read_to_string(state_file) {
+        Err(err) => {
+            if err.kind() != std::io::ErrorKind::NotFound {
+                log::warn!(
+                    "Cannot read initial state from {}: {}",
+                    state_file.display(),
+                    err
+                );
+            }
+            return State::default();
+        }
+        Ok(data) => match serde_json::from_str::<State>(&data) {
+            Err(err) => {
+                log::warn!(
+                    "Cannot parse initial state from {}: {}",
+                    state_file.display(),
+                    err
+                );
+                return State::default();
+            }
+            Ok(state) => {
+                log::debug!("Initial state: {:?}", state);
+                return state;
+            }
+        },
+    }
+}
+
+pub fn store(state_file: &Path, state: &State) {
+    if let Some(parent) = state_file.parent() {
+        if let Err(err) = std::fs::create_dir_all(&parent) {
+            log::warn!(
+                "Cannot create state directory {}: {}",
+                parent.display(),
+                err
+            );
+            return;
+        }
+    }
+
+    let payload = match serde_json::to_string_pretty(&state) {
+        Err(err) => {
+            log::warn!("Cannot serialize state: {}", err);
+            return;
+        }
+
+        Ok(payload) => payload,
+    };
+
+    let write_result = AtomicFile::new(state_file, OverwriteBehavior::AllowOverwrite)
+        .write(|f| f.write_all(payload.as_bytes()));
+    match write_result {
+        Err(err) => log::warn!("Cannot write state to {}: {}", state_file.display(), err),
+        Ok(()) => log::debug!("State stored in {}", state_file.display()),
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use pretty_assertions::assert_eq;
+    use tempfile;
+    use zinnia_runtime::anyhow::Result;
+
+    #[test]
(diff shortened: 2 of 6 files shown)


daemon/main.rs
```diff
@@ -36,6 +38,9 @@ async fn run(config: CliArgs) -> Result<()> {
         ));
     }
 
+    let state_file = PathBuf::from(config.state_root).join("state.json");
+    log::debug!("Using state file: {}", state_file.display());
```
bajtos, line 42: Discussion points:

- Are we fine with using JSON as the serialisation format? A binary format would be more performant, especially considering that we are (re)writing the contents of this file on every job completed.
- The state file will be stored in `STATE_ROOT/state.json`. My intention is to store module-specific states in subdirectories called `STATE_ROOT/{module_name}`, once we get there. I can imagine using `STATE_ROOT/runtime/state.json` for Zinnia state too.

juliangruber: > * Are we fine with using JSON as the serialisation format? A binary format would be more performant, especially considering that we are (re)writing the contents of this file on every job completed.

I assume serialization will be <100ns, which would scale to >10k jobs per second, at which point disk performance has long become the bottleneck already. Therefore I think this is fine.

> * The state file will be stored in `STATE_ROOT/state.json`. My intention is to store module-specific states in subdirectories called `STATE_ROOT/{module_name}`, once we get there. I can imagine using `STATE_ROOT/runtime/state.json` for Zinnia state too.

👍

daemon/state.rs
```diff
@@ -0,0 +1,99 @@
...
+                log::warn!(
+                    "Cannot parse initial state from {}: {}",
+                    state_file.display(),
+                    err
+                );
+                return State::default();
+            }
+            Ok(state) => {
+                log::debug!("Initial state: {:?}", state);
+                return state;
```
juliangruber, line 43:
```suggestion
                State::default()
            }
            Ok(state) => {
                log::debug!("Initial state: {:?}", state);
                state
```

daemon/state.rs
```diff
@@ -0,0 +1,99 @@
...
+            }
+        },
+    }
+}
+
+pub fn store(state_file: &Path, state: &State) {
```
juliangruber, line 49: would this make more sense in the `impl` for `State`?

daemon/state.rs
```diff
@@ -0,0 +1,99 @@
...
+    if let Some(parent) = state_file.parent() {
```
juliangruber, line 50: checking for existing parent dir on every write is expensive, what do you think about we instead let the write fail, and only then create the parent dir?

daemon/state.rs
```diff
@@ -0,0 +1,99 @@
...
+                err
+            );
+            return;
+        }
+    }
+
+    let payload = match serde_json::to_string_pretty(&state) {
+        Err(err) => {
+            log::warn!("Cannot serialize state: {}", err);
+            return;
```
juliangruber, line 64: should we return an error here instead, so that the consumer can handle it (or it can make the process crash)?

daemon/state.rs
```diff
@@ -0,0 +1,99 @@
...
+        }
+
+        Ok(payload) => payload,
+    };
+
+    let write_result = AtomicFile::new(state_file, OverwriteBehavior::AllowOverwrite)
+        .write(|f| f.write_all(payload.as_bytes()));
```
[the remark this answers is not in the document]
juliangruber: Alternative design (using threads):

Let the state update here using an atomic variable, have a separate thread that will periodically flush to disk (eg every 1s).

This also solves the problem of high write count, while not causing any noticeable inconvenience to the user.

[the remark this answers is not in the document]
bajtos: That would work at the risk that we may lose some data if the process exits before we flush to disk. We can mitigate that risk by implementing graceful shutdown (#143), but that's even more complexity.

daemon/state.rs
bajtos: Resolved via 041ac15

daemon/state.rs
bajtos: Great idea!

bajtos: See f256410

daemon/state.rs
bajtos: Great suggestion! See 7e7d49a

daemon/state.rs
bajtos: This serialisation call should not fail unless we make a programmer error. Quoting from `serde_json::to_string_pretty` docs:

```rs
/// Serialization can fail if `T`'s implementation of `Serialize` decides to
/// fail, or if `T` contains a map with non-string keys.
```

If our State becomes impossible to serialise to JSON, then we are pretty much doomed. If we crash the process, then Station should restart it. This will lead to a State that cannot be serialised, so the process crashes again, and Station tries to restart it again. And so on.

By ignoring the error, we can keep running jobs and earning rewards, although the job counter will show incorrect value.

It would be great if `zinniad` could report to us (via Sentry?) when we encounter this problem, but I feel that's beyond the scope of this pull request.

WDYT?

daemon/state.rs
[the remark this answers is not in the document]
juliangruber: What about we just implement the part of graceful shutdown that flushes state to disk?

daemon/state.rs
```diff
@@ -0,0 +1,107 @@
...
+
+        if let Err(atomicwrites::Error::Internal(err)) = &write_result {
+            if err.kind() == std::io::ErrorKind::NotFound {
+                if let Some(parent) = state_file.parent() {
+                    if let Err(err) = std::fs::create_dir_all(&parent) {
+                        log::warn!(
+                            "Cannot create state directory {}: {}",
+                            parent.display(),
+                            err
+                        );
```
juliangruber, line 71: should this crash the process? I think the user would like to be notified that something is fundamentally wrong

daemon/state.rs
juliangruber: > If our State becomes impossible to serialise to JSON, then we are pretty much doomed. If we crash the process, then Station should restart it. This will lead to a State that cannot be serialised, so the process crashes again, and Station tries to restart it again. And so on.

To me that's better than running modules but being fundamentally unable to deal with their results. Station should give up after a while and report there is a series problem. It already will do that, using Sentry.

juliangruber (changes requested):

daemon/state.rs
bajtos: Fair enough. Should I crash the process on all errors, including I/O errors when writing the state file, or only when we cannot serialise to JSON?

juliangruber: I think only on fatal errors that you can't recover from. Failing to write a file once can be recoverable, being unable to serialize state not

daemon/state.rs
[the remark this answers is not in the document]
bajtos: See https://github.com/filecoin-station/zinnia/issues/201

daemon/state.rs
bajtos: In the call, we agreed to crash on all errors for now. I made the changes in 6055e99. I have to say I like the new code much more. Thanks for the nudge to improve this part!

juliangruber (approved):

PR #205, opened 2023-05-10
feat: allow importing module-local JS files

bajtos:
Allow Zinnia modules to import other JavaScript files that are inside the module directory. The module directory is determined as the parent directory of the main module file (the entrypoint).

Out of scope: consuming 3rd-party packages/dependencies.

See #43

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4076,9 +4076,9 @@ checksum = "388a1df253eca08550bef6c72392cfe7c30914bf41df5269b68cbd6ff8f570a3"
 
 [[package]]
 name = "serde"
-version = "1.0.162"
+version = "1.0.163"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "71b2f6e1ab5c2b98c05f0f35b236b22e8df7ead6ffbf51d7808da7f8817e7ab6"
+checksum = "2113ab51b87a539ae008b5c6c02dc020ffa39afd2d83cffcb3f4eb2722cebec2"
 dependencies = [
  "serde_derive",
 ]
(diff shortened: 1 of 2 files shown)


docs/building-modules.md
```diff
@@ -30,6 +30,51 @@ Hello universe!
 
 See [example modules](../examples) for more advanced examples.
 
+## Modules
+
+Zinnia supports ES Modules (also known as
+[JavaScript Modules](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Modules)).
+
+For sandboxing reasons, module imports are limited to files in the root directory of the Zinnia
+module being executed.
```
bajtos, line 39: Why I choose this design:

- We can stay somewhat unopinionated about the file naming conventions in Zinnia modules.
- We are imposing only one small restriction that the module entry point must be in the root directory. 
- Developers can choose whatever file name they like - `main.js`, `index.js`, `spark.js`, etc.
- This design works both for `zinnia` CLI and `zinniad` inside Filecoin Station.

juliangruber:
> A more secure solution would involve resolving the imported path into the real target and applying the path-based filter on that target. I think that would be an overkill right now.

Could we also prevent symlink resolution for now?

docs/building-modules.md
```diff
@@ -30,6 +30,51 @@ Hello universe!
...
+```js
+// These imports are allowed in `main.js`
+import { processJob } from "./lib.js";
+import { log } from "./util/log.js";
+
+// This will be rejected at runtime
+import * as code from "../other/code.js";
+
+// This will work in `util/log.js`
+import { format } from "../lib.js";
```
juliangruber, line 72: great that you added an example for a valid parent directory immport

runtime/tests/js/module_fixtures/lib.js
```diff
@@ -0,0 +1,2 @@
+// Verify that static imports work
+export { log } from "./log.js";
```
juliangruber, line 2: awesome that you added tests too! <3

juliangruber (approved):
SHIPIT

PR #212, opened 2023-05-15
feat: test runner

bajtos:
_work in progress_

### DIFF
--- a/runtime/tests/js/helpers.js
+++ b/runtime/tests/js/helpers.js
@@ -1,23 +1,67 @@
 // A dummy wrapper to create isolated scopes for individual tests
 // We should eventually replace this with a proper test runner
 // See https://github.com/filecoin-station/zinnia/issues/30
-export function test(name, fn) {
+let globalRoot;
+function getGlobalRoot() {
+  if (!globalRoot) {
+    globalRoot = {
+      pendingTests: [],
+      result: {
+        passed: 0,
+        failed: [],
+      },
+    };
+    Promise.resolve().then(runNextTest);
+  }
+  return globalRoot;
+}
+
+function runNextTest() {
+  const nextTest = globalRoot.pendingTests.shift();
+
+  if (nextTest) {
+    Promise.resolve(executeTest(nextTest)).then(runNextTest);
+    return;
+  }
+
+  for (const err of globalRoot.result.failed) {
+    console.error(err);
+    console.error("\n");
+  }
+
+  console.log("\nPASSED: %s FAILED: %s", globalRoot.result.passed, globalRoot.result.failed.length);
+  return;
+}
+
+function executeTest({ name, fn }) {
   let maybePromise;
   try {
     maybePromise = fn();
   } catch {
     err.message = `Test ${name} failed. ` + err.message;
-    throw err;
+    globalRoot.result.failed.push(err);
+    return;
   }
 
   if (typeof maybePromise?.then !== "function") {
     // The test fn was synchronous, we are done.
+    globalRoot.result.passed++;
     return maybePromise;
   }
 
   // The test fn returned a promise, we need to wait for completion
-  return maybePromise.catch((err) => {
-    err.message = `Test ${name} failed. ` + err.message;
-    throw err;
-  });
+  return maybePromise.then(
+    (ok) => {
+      globalRoot.result.passed++;
+      return ok;
+    },
+    (err) => {
+      err.message = `Test ${name} failed. ` + err.message;
+      getGlobalRoot().result.failed.push(err);
+    },
+  );
+}
+
+export function test(name, fn) {
+  getGlobalRoot().pendingTests.push({ name, fn });
 }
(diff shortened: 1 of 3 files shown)


bajtos:
Screenshot showing how we run 7 tests, and one of the tests fail:

Screenshot 2023-05-15 at 17 18 26

bajtos:
I added more features (e.g. a test reporter) and cleaned up the code. The PR is ready for review & landing.

Screenshot from running tests via `zinnia` CLI:

Screenshot 2023-05-16 at 15 35 16

Screenshot from Zinnia's `cargo test`:

Screenshot 2023-05-16 at 15 39 06

runtime/js/test.js
```diff
@@ -0,0 +1,209 @@
...
+
+  if (nextTest) {
+    nextTest.execute(nextTest).then(runNextTest);
+    return;
+  }
+
+  reportTestResults();
+}
+
+function reportTestResults() {
```
bajtos, line 141: Discussion point.

There are no tests for this test runner and the test reporter. We have a certain amount of implicit code coverage thanks to running Zinnia tests using this same test runner. Is it enough, though?

Should I add a few tests to verify the most important scenarios, including tests checking the output printed to stdout & stderr?

runtime/ext.rs
```diff
@@ -48,6 +48,8 @@ deno_core::extension!(
       "06_util.js",
       "90_zinnia_apis.js",
       "98_global_scope.js",
+      "internals.js",
+      "test.js",
```
bajtos, line 52: FWIW, we are always bundling and loading our test framework, even when running a module inside Filecoin Station via `zinniad`.

While it's not ideal, I think it's acceptable for the current state of our project.

runtime/runtime.rs
```diff
@@ -157,6 +157,9 @@ impl ModuleLoader for ZinniaModuleLoader {
         referrer: &str,
         _kind: ResolutionKind,
     ) -> Result<ModuleSpecifier, AnyError> {
+        if specifier == "zinnia:test" {
+            return Ok(ModuleSpecifier::parse("ext:zinnia_runtime/test.js").unwrap());
```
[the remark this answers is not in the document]
bajtos: The downside: error stack traces now show `ext:zinnia_runtime/test.js` instead of `zinnia:test` as the filename 🤷🏻

runtime/js/test.js
```diff
@@ -0,0 +1,209 @@
...
+      const err = new Error();
+      err.name = "[some test(s) failed]";
+      err.message = "";
+      err.stack = "";
+      throw err;
+    });
+  }
+}
+
+export function test(name, fn) {
```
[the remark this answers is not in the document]
juliangruber: I don't think so, we'd end up with a potentially confusing half implementation of BDD. I think it's safer either to fully support the expected interface, or not to claim it at all

runtime/js/test.js
juliangruber: Output checking sgtm! This is also the primary way the node core test library is tested

juliangruber: If we don't have that and only the "living spec", then it will be tricky to say if an error is because of a test bug or a runner bug

runtime/js/test.js
```diff
@@ -0,0 +1,211 @@
...
+  console.log(
+    "\n%s | %s passed | %s failed %s",
+    failed ? red("FAIL") : green("ok"),
+    passed,
+    failed,
+    grey(`(${duration}ms)`),
+  );
+  // Signal test failure by creating an unhandled error,
+  // so that `zinnia` CLI returns a non-zero exit code.
+  // TODO(bajtos) Find a more elegant solution, e.g. add es areI for setting the exit code
```
juliangruber, line 163:
```suggestion
  // TODO(bajtos) Find a more elegant solution, e.g. add an API for setting the exit code
```

I assume you meant this?

runtime/js/test.js
```diff
@@ -0,0 +1,211 @@
...
+  if (failed) {
+    setTimeout(() => {
```
juliangruber, line 165:
```suggestion
    Promise.resolve().then(() => {
```

Would this be more consistent with the rest of the lib?

runtime/runtime.rs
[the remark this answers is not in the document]
juliangruber: Let's fix the stack traces later 👍

juliangruber (changes requested):

runtime/js/test.js
bajtos: Not really. The message printed for unhandled rejections is different from the message printed for uncaught errors.

My version prints this:

```
error: Uncaught [some test(s) failed]
```


Your suggestion leads to this:

```
error: Uncaught (in promise) [some test(s) failed]
```

runtime/js/test.js
bajtos: Makes sense. Adding tests for the test runner's output required more work on our testing infrastructure, but it was worth it. See 7a29089.

juliangruber (approved):
so cool

Issue #44, opened 2023-02-13
Automated tests for JS modules


bajtos:
Make it easy for module authors to write automated tests for their modules.

We need this functionality too for writing our internal tests for Zinnia JS APIs.

See https://deno.land/manual@v1.30.3/basics/testing

## Features to consider:

- framework (API to define tests, test runner, test reporter)
  - collect test files
  - register all test suites (`describe`) and tests (`it`)
  - execute the tests
  - report test results
- CLI, e.g `zinnia test`
- Documentation for module authors.
- Unit-testing vs integration testing.
- Usability features
  - Run only a subset of tests
    - All tests in a single file (`mocha test/file.js`)
    - All tests matching a pattern (`mocha -g regex`)
    - Skip tests (`it.skip` and `describe.skip`)
    - Run only tests selected in code (`it.only`)

See https://deno.land/manual@v1.30.3/basics/testing for inspiration.

A Mastodon thread about DX aspects of JS testing frameworks: https://mastodon.social/@jamesshore@mastodon.online/110345771809205197

## Out of scope

- ~~Assertion library - people can bundle & vendor e.g. https://deno.land/std@0.177.0/testing/asserts.ts~~
- Code coverage
- https://github.com/filecoin-station/zinnia/issues/215
- https://github.com/filecoin-station/zinnia/issues/214

## Subtasks

- [x] #212 
- [x] Relax import sandboxing to allow a nested test file to access all production files
      https://github.com/filecoin-station/zinnia/pull/218
      https://github.com/filecoin-station/zinnia/pull/220
- [x] Embed `assert` module again, this time as `zinnia:assert`
      https://github.com/filecoin-station/zinnia/pull/217
- [x] Write documentation for module authors
      https://github.com/filecoin-station/zinnia/pull/222
- [x] Use `format_test_error` from Deno's `cli/tools.rs` to get nicer error stack traces
      https://github.com/filecoin-station/zinnia/pull/223

Estimate: Time-boxed to 2-3 days

PR #217, opened 2023-05-17
feat: embed Deno assertions as `zinnia:assert`

bajtos:
Example usage:

```js
import { test } from "zinnia:test";
import { assertEquals } from "zinnia:assert";

test("AbortController", () => {
  assertEquals(typeof AbortController, "function", "typeof AbortController");
  assertEquals(AbortController.name, "AbortController", "AbortController.name");
});
```

See #44

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -3,7 +3,7 @@
 !**/*.md
 !**/*.js
 runtime/vendored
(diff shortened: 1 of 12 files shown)


juliangruber (approved):
👌

PR #218, opened 2023-05-17
feat: `zinnia` CLI can import JS from anywhere

bajtos:
Relax the restriction preventing Zinnia modules to load files only under the directory of the main entry point. This enables developers to run individual test files regardless of where they are in the project tree, while allowing the test files to import files from anywhere in the project.

The changes are affecting `zinnia` CLI only. When running inside Filecoin Station, `zinniad` keeps enforcing the sandbox.

See #44

### DIFF
--- a/cli/main.rs
+++ b/cli/main.rs
@@ -1,12 +1,15 @@
 mod args;
 
+use std::rc::Rc;
+use std::time::Duration;
+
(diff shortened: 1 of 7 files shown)


juliangruber (approved):
NOICE

PR #220, opened 2023-05-18
refactor: cleanup code handling import sandboxing

bajtos:
This is a follow-up for #218 which I accidentally merged too early.

### DIFF
--- a/runtime/module_loader.rs
+++ b/runtime/module_loader.rs
@@ -20,8 +20,14 @@ pub struct ZinniaModuleLoader {
 }
 
 impl ZinniaModuleLoader {
-    pub fn new(module_root: Option<PathBuf>) -> Self {
-        Self { module_root }
+    pub fn build(module_root: Option<PathBuf>) -> Result<Self> {
+        let module_root = match module_root {
+            None => None,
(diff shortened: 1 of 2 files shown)


runtime/module_loader.rs
```diff
@@ -20,8 +20,14 @@ pub struct ZinniaModuleLoader {
 }
 
 impl ZinniaModuleLoader {
-    pub fn new(module_root: Option<PathBuf>) -> Self {
-        Self { module_root }
+    pub fn build(module_root: Option<PathBuf>) -> Result<Self> {
+        let module_root = match module_root {
+            None => None,
+            // We must canonicalize the module root path too. It's best to do it once at startup.
+            Some(r) => Some(r.canonicalize()?),
```
bajtos, line 27: This change has an important side effect: when `module_root` points to a directory that does not exist then `ZinniaModuleLoader::build` returns an error now. In other words, we fail at startup.

Before this change, when we were not able to canonicalize the root directory, then we would treat all module imports as outside of the project root, and therefore reject them.

juliangruber: sounds great to me!

PR #222, opened 2023-05-18
docs: testing guide for module authors

bajtos:
See #44

### DIFF
--- a/docs/building-modules.md
+++ b/docs/building-modules.md
@@ -1,4 +1,4 @@
-# Building modules
+# Building Modules
 
 A Station Module is a long-running process that's performing jobs like network probes, content
 delivery, and computation.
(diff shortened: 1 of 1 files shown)


docs/building-modules.md
```diff
@@ -30,7 +30,13 @@ Hello universe!
 
 See [example modules](../examples) for more advanced examples.
 
-## Modules
+## Table of Contents
+
+- [Importing JavaScript Modules](#importing-javascript-modules)
+- [Platform APIs](#platform-apis)
+- [Testing Guide](#testing-guide)
```
bajtos, line 37: GitHub provides a built-in table of contents. However, I think may people are not aware of that feature. I am adding minimalistic ToCs here and there to make this page easier to navigate.

juliangruber: Thanks for sharing, I wasn't aware of this either

PR #224, opened 2023-05-18
fix: user agent reported by Fetch client

### LINKED ISSUE #76: Empty user agent (Fetch API)
When making Fetch API request, zinnia should report the runtime version and module name in the user agent.

IMO, the user agent string should be hard-coded, and modules should not be able to change it. At least not until we have very good reasons to allow them to do so.

At the moment, the user agent header is empty:

```json
{
  "accept": "*/*",
  "accept-language": "*",
  "user-agent": "",
  "accept-encoding": "gzip, br",
  "host": "localhost:3000"
}
```

Echo HTTP server in Node.js:

```js
require('http').createServer((req, res) => {
  res.write(JSON.stringify(req.headers, null, 2));
  res.end();
}).listen(3000)
```

Zinnia module making the request and producing the output shown at the top:

```js
console.log(
  await (await fetch('http://localhost:3000/')).text()
);
```

Estimate: 1 day

bajtos:
Change the value sent in `User-Agent` request headers to match the agent version used to identify our libp2p node.

Modify the agent string as follows:

- CLI keeps reporting `zinnia/0.9.0` (`0.9.0` is the current version)

- Zinniad (Filecoin Station) reports the user agent as a combination of the Zinnia version plus the name of the Station module running.

  Example: `zinniad/0.9.0 dummy-module`

Resolves #76

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5751,6 +5751,7 @@ dependencies = [
 name = "zinnia_runtime"
 version = "0.9.0"
 dependencies = [
+ "assert_fs",
  "atty",
  "chrono",
  "console_static_text",
--- a/daemon/main.rs
+++ b/daemon/main.rs
@@ -48,7 +48,6 @@ async fn run(config: CliArgs) -> Result<()> {
     // TODO: configurable module name and version
     // https://github.com/filecoin-station/zinnia/issues/147
     let module_name = file.trim_end_matches(".js");
-    let module_version = "unknown";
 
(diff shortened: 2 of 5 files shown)


bajtos:
I would like to add a test to verify that we are correctly configuring the user agent. The test setup is non-trivial, as I need to create some sort of an HTTP server inside our Rust test codebase.

juliangruber (approved):

PR #230, opened 2023-05-25
feat: print initial job count on `zinniad` start

bajtos:
Announce the total number of completed jobs immediately at startup.

Before this change, we would not announce it until the first job was completed. That led to suboptimal UX: the Station initially displayed job count 0. After a while, the counter jumped to a much higher value.

With this change in place, Filecoin Station should have the right value of our job counter as soon as we load it from our persisted state.

See https://github.com/filecoin-station/core/pull/134#discussion_r1203790934

### DIFF
--- a/daemon/station_reporter.rs
+++ b/daemon/station_reporter.rs
@@ -28,7 +28,7 @@ impl StationReporter {
             .unwrap()
             .total_jobs_completed;
 
-        Self {
+        let reporter = Self {
             tracker: RefCell::new(JobCompletionTracker::new(
                 initial_job_count,
                 job_report_delay,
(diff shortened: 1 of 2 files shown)


daemon/station_reporter.rs
```diff
@@ -28,15 +28,23 @@ impl StationReporter {
...
             module_name,
             log_target,
             state_file,
+        };
+
+        // Report the initial job count to prevent Station Desktop from showing incorrect job count
+        // until a Zinnia module completes the first job
+        if initial_job_count > 0 {
+            reporter.print_jobs_completed(initial_job_count);
         }
```
juliangruber, line 45:
```suggestion
        reporter.print_jobs_completed(initial_job_count);
```

This would be even better for Desktop, because it will know that that module's job count isn't loading any more, even when it's zero

juliangruber (approved):

PR #243, opened 2023-06-07
feat: IPFS retrieval client

bajtos:
_details TBD_

See https://github.com/filecoin-station/roadmap/issues/19

### DIFF
--- a/cli/main.rs
+++ b/cli/main.rs
@@ -1,6 +1,7 @@
 mod args;
 
 use std::rc::Rc;
+use std::sync::Arc;
 use std::time::Duration;
 
 use args::{CliArgs, Commands};
@@ -9,7 +10,9 @@ use clap::Parser;
 use zinnia_runtime::anyhow::{Context, Error, Result};
 use zinnia_runtime::deno_core::error::JsError;
 use zinnia_runtime::fmt_errors::format_js_error;
-use zinnia_runtime::{colors, resolve_path, run_js_module, BootstrapOptions, ConsoleReporter};
+use zinnia_runtime::{
+    colors, lassie, resolve_path, run_js_module, BootstrapOptions, ConsoleReporter,
+};
 
 #[tokio::main(flavor = "current_thread")]
 async fn main() {
@@ -32,9 +35,19 @@ async fn main_impl() -> Result<()> {
                 &file,
                 &std::env::current_dir().context("unable to get current working directory")?,
             )?;
+
+            let lassie_daemon = Arc::new(
+                lassie::Daemon::start(lassie::DaemonConfig {
+                    temp_dir: None, // TODO: Should we use something like ~/.cache/zinnia/lassie?
+                    port: 0,
+                })
+                .context("cannot initialize the IPFS retrieval client Lassie")?,
+            );
+
             let config = BootstrapOptions::new(
                 format!("zinnia/{}", env!("CARGO_PKG_VERSION")),
                 Rc::new(ConsoleReporter::new(Duration::from_millis(500))),
+                lassie_daemon,
                 None,
             );
             run_js_module(&main_module, &config).await?;
--- a/runtime/Cargo.toml
+++ b/runtime/Cargo.toml
@@ -21,6 +21,8 @@ deno_fetch = "0.129.0"
 deno_url = "0.105.0"
 deno_web = "0.136.0"
 deno_webidl = "0.105.0"
+# lassie = "0.1.1"
+lassie = { git = "https://github.com/filecoin-station/rusty-lassie.git" }
 log.workspace = true
 once_cell = "1.18.0"
 serde.workspace = true
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2179,6 +2179,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "lassie"
+version = "0.1.1"
+source = "git+https://github.com/filecoin-station/rusty-lassie.git#0980b5c08c3dbd92eb658caf34649dc99d554798"
+dependencies = [
+ "cc",
+ "log",
+]
+
+[[package]]
 name = "lazy-regex"
 version = "2.5.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -5682,6 +5691,7 @@ dependencies = [
  "deno_web",
  "deno_webidl",
  "env_logger",
+ "lassie",
  "log",
  "once_cell",
  "pretty_assertions",
--- a/daemon/main.rs
+++ b/daemon/main.rs
@@ -4,14 +4,14 @@ mod station_reporter;
 
 use std::path::PathBuf;
 use std::rc::Rc;
+use std::sync::Arc;
 use std::time::Duration;
 
 use args::CliArgs;
 use clap::Parser;
 
-use log::{error, info};
 use zinnia_runtime::anyhow::{anyhow, Context, Error, Result};
-use zinnia_runtime::{get_module_root, resolve_path, run_js_module, BootstrapOptions};
+use zinnia_runtime::{get_module_root, lassie, resolve_path, run_js_module, BootstrapOptions};
 
 use crate::station_reporter::{log_info_activity, StationReporter};
 
@@ -27,7 +27,7 @@ async fn main() {
 }
 
 async fn run(config: CliArgs) -> Result<()> {
-    info!("Starting zinniad with config {config:?}");
+    log::info!("Starting zinniad with config {config:?}");
 
     if config.files.is_empty() {
         return Err(anyhow!("You must provide at least one module to run."));
@@ -40,6 +40,15 @@ async fn run(config: CliArgs) -> Result<()> {
 
     let state_file = PathBuf::from(config.state_root).join("state.json");
     log::debug!("Using state file: {}", state_file.display());
+
+    let lassie_config = lassie::DaemonConfig {
+        temp_dir: Some(PathBuf::from(config.cache_root).join("lassie")),
+        port: 0,
+    };
+    let lassie_daemon = Arc::new(
+        lassie::Daemon::start(lassie_config)
+            .context("cannot initialize the IPFS retrieval client Lassie")?,
+    );
 
     log_info_activity("Module Runtime started.");
 
@@ -63,6 +72,7 @@ async fn run(config: CliArgs) -> Result<()> {
             Duration::from_millis(200),
             module_name.into(),
         )),
+        lassie_daemon,
         module_root: Some(module_root),
         no_color: true,
         is_tty: false,
@@ -88,6 +98,6 @@ fn exit_with_error(error: Error) {
     let error_string = format!("{error:?}");
     let error_code = 1;
 
-    error!("{}", error_string.trim_start_matches("error: "));
+    log::error!("{}", error_string.trim_start_matches("error: "));
     std::process::exit(error_code);
 }
--- a/runtime/lib.rs
+++ b/runtime/lib.rs
@@ -18,4 +18,6 @@ mod reporter;
 pub use console_reporter::*;
 pub use reporter::*;
 
+pub use lassie;
+
 mod ext;
--- a/runtime/runtime.rs
+++ b/runtime/runtime.rs
@@ -1,5 +1,6 @@
 use std::path::PathBuf;
 use std::rc::Rc;
+use std::sync::Arc;
 
 use deno_core::{located_script_name, serde_json, JsRuntime, ModuleSpecifier, RuntimeOptions};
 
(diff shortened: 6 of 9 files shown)


bajtos:
> Are we happy with `fetch('ipfs://bafycid')` API? I believe this API is provided by Brave when the IPFS integration is enabled.

I did a test in Brave:
1. Enable IPFS integration in settings
2. Load `ipfs://bafybeib36krhffuh3cupjml4re2wfxldredkir5wti3dttulyemre7xkni` (otherwise `ipfs://` requests are blocked by the built-in CORS rules)
3. Open Dev Tools > Console
4. Run the following code:
    ```js
    await (await fetch(
      "ipfs://bafybeib36krhffuh3cupjml4re2wfxldredkir5wti3dttulyemre7xkni"
    )).text()
    ```
5. We get back the content of the file (not CAR):

   ```
   My most famous drawing, and one of the first I did for the site
   ```

    Screenshot 2023-06-08 at 17 30 41

6. Append `?format=car` to the URL to get back the CAR file

    ```js
    await (await fetch(
      "ipfs://bafybeib36krhffuh3cupjml4re2wfxldredkir5wti3dttulyemre7xkni?format=car"
    )).arrayBuffer()
    ```
   
    Screenshot 2023-06-08 at 17 35 06

My conclusion is that IPFS Gateways and Brave's built-in IPFS node (presumably based on Kubo) has different default behaviour from Lassie:
- Lassie defaults to CAR
- Kubo & GW default to autodetect the file's content type and return the file's content (decoded from CAR)

If we feel this can be confusing to Zinnia users, then I am proposing to enforce modules to explicitly include either `?format=car` in the query string or set `Accept: application/vnd.ipld.car` request header when fetching `ipfs://` URLs. This allows us to implement GW-like behaviour without breaking existing code.

Having said that, considering the early alpha status of Zinnia, I think we can afford to introduce breaking changes, and it's better to invest our time in things adding more value.

@juliangruber WDYT?

bajtos:
Reminder for myself: after we land this PR, I need to update our Windows release builds to include golassie.dll in the zip archives.

runtime/Cargo.toml
```diff
@@ -21,6 +21,8 @@ deno_fetch = "0.129.0"
 deno_url = "0.105.0"
 deno_web = "0.136.0"
 deno_webidl = "0.105.0"
+lassie = "0.3.0"
```
mxinden, line 24: Neat. I didn't know there is a Rust client!

cli/main.rs
```diff
@@ -32,9 +35,19 @@ async fn main_impl() -> Result<()> {
                 &file,
                 &std::env::current_dir().context("unable to get current working directory")?,
             )?;
+
+            let lassie_daemon = Arc::new(
+                lassie::Daemon::start(lassie::DaemonConfig {
+                    temp_dir: None, // TODO: Should we use something like ~/.cache/zinnia/lassie?
```
juliangruber, line 41: Station Core will pass `$CACHE_ROOT`:

https://github.com/filecoin-station/core/blob/7f07e5203c71366fa9bde713b0d28dee9ea0c51d/lib/zinnia.js#L48-L55.

Can we make Zinnia use that if it is set?

docs/building-modules.md
```diff
@@ -329,6 +330,32 @@ Report that a single job was completed.
 
 Call this function every time your module completes a job. It's ok to call it frequently.
 
+### IPFS Retrieval Client
+
+Zinnia provides a built-in IPFS retrieval client making it easy to fetch content-addressed data from
+IPFS and Filecoin networks. You can retrieve data for a given CID using the web platform API `fetch`
+and using the URL scheme `ipfs://`.
```
juliangruber, line 337:
```suggestion
IPFS and Filecoin networks. You can retrieve data for a given CID using the web platform API `fetch`
together with the URL scheme `ipfs://`.
```

docs/building-modules.md
```diff
@@ -329,6 +330,32 @@ Report that a single job was completed.
...
+Example:
+
+```js
+const response = await fetch("ipfs://bafybeib36krhffuh3cupjml4re2wfxldredkir5wti3dttulyemre7xkni");
+assert(response.ok);
+const data = await response.arrayBuffer();
+// data contains binary data in the CAR format
+```
+
+> Note: At the moment, Zinnia does not provide any tools to interpret the returned CAR data. We are
```
juliangruber, line 348:
```suggestion
> Note: At the moment, Zinnia does not provide any tools for interpreting the returned CAR data. We are
```

docs/building-modules.md
```diff
@@ -329,6 +330,32 @@ Report that a single job was completed.
...
+> discussing support for reading UnixFS data in
+> [zinnia#245](https://github.com/filecoin-station/zinnia/issues/246).
+
+Under the hood, Zinnia handles `ipfs://bafy...` requests by calling Lassie HTTP API. You can learn
```
juliangruber, line 352:
```suggestion
Under the hood, Zinnia handles `ipfs://bafy...` requests by calling Lassie's HTTP API. You can learn
```

runtime/js/fetch.js
```diff
@@ -0,0 +1,66 @@
+import { fetch as fetchImpl } from "ext:deno_fetch/26_fetch.js";
+import { fromInnerResponse, toInnerResponse } from "ext:deno_fetch/23_response.js";
+import { toInnerRequest, fromInnerRequest, Request } from "ext:deno_fetch/23_request.js";
+import { guardFromHeaders } from "ext:deno_fetch/20_headers.js";
+
+let ipfsScheme = "ipfs://";
```
juliangruber, line 6:
```suggestion
const ipfsScheme = "ipfs://";
```

I believe it doesn't need to be mutable

runtime/js/fetch.js
```diff
@@ -0,0 +1,66 @@
...
+let ipfsBaseUrl = undefined;
+
+export function setLassieUrl(/** @type {string} */ value) {
+  ipfsBaseUrl = value + "ipfs/";
+}
+
+export function fetch(resource, options) {
+  let request = new Request(resource, options);
+  // Fortunately, Request#url is a string, not an instance of URL class
```
juliangruber, line 15: What is fortunate about that?

runtime/tests/js/ipfs_retrieval_tests.js
```diff
@@ -0,0 +1,53 @@
...
+const EXPECTED_CAR_BASE64 =
+  "OqJlcm9vdHOB2CpYJQABcBIgO/KicpaH2Kj0sXyJNWLdY4kGpEe2mjY5zovBGRJ+6mpndmVyc2lvbgFrAXASIDvyonKWh9io9LF8iTVi3WOJBqRHtpo2Oc6LwRkSfupqCkUIAhI/TXkgbW9zdCBmYW1vdXMgZHJhd2luZywgYW5kIG9uZSBvZiB0aGUgZmlyc3QgSSBkaWQgZm9yIHRoZSBzaXRlGD8=";
+
+test("can retrieve CID content as a CAR file", async () => {
+  const requestUrl = "ipfs://bafybeib36krhffuh3cupjml4re2wfxldredkir5wti3dttulyemre7xkni";
+
+  const response = await fetch(requestUrl);
+  if (!response.ok) {
+    throw new AssertionError(
+      `Fetch request failed with status code ${response.status}: ${await response.body()}`,
```
juliangruber, line 13:
```suggestion
      `Fetch request failed with status code ${response.status}: ${await response.text()}`,
```

https://developer.mozilla.org/en-US/docs/Web/API/Response/body

Looks like you didn't make this test fail :P (ref to our call about testing best practices)

runtime/tests/js/ipfs_retrieval_tests.js
juliangruber, line 1: I think we haven't yet tested what happens when there's an issue with the Lassie response. Do you have any ideas how to test that?

juliangruber (changes requested):
>  Are we happy with fetch('ipfs://bafycid') API? I believe this API is provided by Brave when the IPFS integration is enabled.

It's intuitive and already established, I'd also say let's keep it until we require a different strategy (which could be added in addition to the always nice fetch() API)

> Do we want to hide any of the Lassie response headers from Station Modules making the retrieval requests?

I'm not sure whether keeping or removing has advantages. Let's see what happens?

juliangruber:
> If we feel this can be confusing to Zinnia users, then I am proposing to enforce modules to explicitly include either ?format=car in the query string or set Accept: application/vnd.ipld.car request header when fetching ipfs:// URLs. This allows us to implement GW-like behaviour without breaking existing code.
>
> Having said that, considering the early alpha status of Zinnia, I think we can afford to introduce breaking changes, and it's better to invest our time in things adding more value.

I couldn't agree more!

runtime/Cargo.toml
bajtos: It's a thin Rust wrapper embedding the original Go Lassie, I started the project three weeks ago :)

https://github.com/filecoin-station/rusty-lassie

cli/main.rs
bajtos: I am already using CACHE_ROOT in `zinniad`, see here:

https://github.com/filecoin-station/zinnia/pull/243/files#diff-c74dac62db9c80f1be22978c93249f7b304e05ddb38131e7969efa13effaeb1eR45

The file `cli/main.js` implements `zinnia`, the CLI people use locally when building Station modules. Let's explore together what a good developer experience would look like?

IMO:
- We should not force `zinnia` users to always provide `CACHE_ROOT`. We don't ask them for `FIL_WALLET_ADDRESS` either. This way, users can type `zinnia run main.js`, and all works out of the box.
- I guess allowing CLI users to control the CACHE ROOT can be helpful. I am not sure, though, if an env var provides good ergonomy. Would a project-specific config file be easier to use?
- How important is this? Can we leave the current solution and open a follow-up GH issue to discuss what would a good (and easy-to-implement) solution look like?
- Note: if we tell Lassie to use a specific temp dir that's not automatically cleaned by the operating system, we will need to clean any leftover files ourselves.

runtime/js/fetch.js
bajtos: Thank you for catching this. I often forget myself and write my JavaScript code in Rust style these days 🙈

cli/main.rs
bajtos: > Note: if we tell Lassie to use a specific temp dir that's not automatically cleaned by the operating system, we will need to clean any leftover files ourselves.

I'll be implementing this cleanup in `zinnia` as part of https://github.com/filecoin-station/zinnia/issues/245

runtime/tests/js/ipfs_retrieval_tests.js
juliangruber: I guess if you had started with the test and then added the fetch implementation, that AssertionError would have been tested. For now I think just fixing the calls is enough. If you wanted you could refactor a test utility like

```js
const assertResponseOk = () => { ... }
```

cli/main.rs
juliangruber: I wasn't suggesting always needing to pass `CACHE_ROOT`, I thought if it's not passed then lassie shall pick its own temp dir, if it is passed (as in Station Core) it shall use that, just to keep all of the files together.

The primary use case for changing lassie's temp dir to me is not CLI usage but inside Station.

> Note: if we tell Lassie to use a specific temp dir that's not automatically cleaned by the operating system, we will need to clean any leftover files ourselves.

That's a great point! What does Lassie use by default rn? If it uses an OS cleaned up dir, I'd suggest to leave it at that and add a comment summarizing our discussion here

runtime/tests/js/ipfs_retrieval_tests.js
bajtos: Added a helper and verified that it works as intended:

```
ipfs_retrieval_tests.js
  ✖ can retrieve CID content as a CAR file (5674ms)
Error: AssertionError: Fetch request failed with status code 404: no candidates found

    at assertResponseIsOk (file:///Users/bajtos/src/zinnia/zinnia/runtime/tests/js/ipfs_retrieval_tests.js:48:11)
    at eventLoopTick (ext:core/01_core.js:181:11)
```

runtime/js/fetch.js
bajtos: I decided to write a longer comment, see 7db4ee8

```js
  // The `resource` arg can be a string or any other object with a stringifier — including a URL
  // object — that provides the URL of the resource you want to fetch; or a Request object.
  // See https://developer.mozilla.org/en-US/docs/Web/API/fetch#parameters
  // Fortunately, Request's constructor handles the conversions, and Request#url is always a string.
  // See https://developer.mozilla.org/en-US/docs/Web/API/Request/url
```

cli/main.rs
bajtos: > That's a great point! What does Lassie use by default rn? If it uses an OS cleaned up dir, I'd suggest to leave it at that and add a comment summarizing our discussion here

Quoting from Lassie comments:
https://github.com/filecoin-project/lassie/blob/afc2ee5a4bc6f5e22ef2cc69396cc9b25f57b854/pkg/lassie/lassie.go#L199-L201
```go
// WithTempDir allows you to specify a custom temp directory for bitswap
// retrievals, used for a temporary block store for the preloader. The default
// is the system temp directory.
```

I think that should be good enough for now, even if we may end up leaving some temporary files behind when `zinnia` exists unexpectedly.

juliangruber (approved):

PR #257, opened 2023-06-21
fix: mkdirp lassie `temp_dir`

juliangruber:


daemon/main.rs
```diff
@@ -40,9 +41,11 @@ async fn run(config: CliArgs) -> Result<()> {
 
     let state_file = PathBuf::from(config.state_root).join("state.json");
     log::debug!("Using state file: {}", state_file.display());
+    let temp_dir = PathBuf::from(config.cache_root).join("lassie");
```
bajtos, line 44: Let's call this `lassie_temp_dir`, please, to clarify that we should not use this temp dir for any other components.

bajtos (approved):

bajtos (commented):
:shipit:

PR #266, opened 2023-06-26
fix: relax timer test timeout

juliangruber:
https://github.com/filecoin-station/zinnia/actions/runs/5374177743/jobs/9749361834?pr=264

> Error: Error: setTimeout(50) should take between 40 to 120 ms to execute, but took 208 ms instead

### DIFF
--- a/runtime/tests/js/timers_tests.js
+++ b/runtime/tests/js/timers_tests.js
@@ -6,7 +6,7 @@ await new Promise((resolve) => {
 
 const duration = Date.now() - start;
 const min = 40;
-const max = 120;
+const max = 300;
 
 if (duration < min || duration > max) {
   throw new Error(


bajtos (approved):
It makes me wonder whether this test has any value by now. Asserting that a timeout of 50 ms takes between 40 to 300 ms isn't very useful. But then I guess it's better than no test at all 🤷🏻‍♂️ 

Thank you, @juliangruber, for taking the time to make the test more robust!

juliangruber:
I wondered about this as well. Since it's still legal for the function to take longer time, shall we remove the upper bound and let the test succeed as soon as it calls back?

juliangruber:
Or have a out of reach upper bound, like 10s

bajtos:
IDK. Can we rely on Deno to implement this function correctly? I guess then we can remove the upper bound check. If the timeout does not fire at all, then the entire test will time out.

PR #270, opened 2023-06-26
docs: mention Rusty Lassie in README

bajtos:


README.md
```diff
@@ -54,7 +54,7 @@ See [Building Modules](./docs/building-modules.md).
   [Tokio](https://tokio.rs)
 - **Networking**: [deno_fetch](https://crates.io/crates/deno_fetch),
   [rust-libp2p](https://github.com/libp2p/rust-libp2p)
-- **IPFS:** _TBD_
+- **IPFS:** [Rusty Lassie](https://github.com/filecoin-station/rusty-lassie) for retrievals. _TBD: Content providing_
```
juliangruber, line 57: What do you mean by "Content providing"?

bajtos: Allow other IPFS peers to retrieve content produced by a Station Module.

For example, if we wanted to implement a Bacalhau-like worker processing some input CIDs, saving the outputs as content-addressed data and submitting the CID of the output to the orchestrator (or a smart contract), then we must announce the output CID and our routing info to DHT/IPNI/... and allow other nodes to retrieve the content from the Station using Bitswap/Graphsync/HTTP. 

There are other ways how to implement this too. If we don't want Stations to serve the content, we can upload it to a pinning service like web3.storage.

How about the following alternatives?

- TBD: Allow other peers to retrieve content from us
- TBD: Publish content to IPFS

juliangruber: Aaaah I see! I thought this was about returning parsed CAR content from the `fetch()` calls.

Either of your suggestions works for me :)

PR #273, opened 2023-06-28
feat: secure Lassie server using authorization

bajtos:
Configure Lassie HTTP server to require authorization with the configured access token.

This change does not affect Zinnia modules with one exception: `fetch('ipfs://`) requests are not allowed to send an `Authorization` header now.

Also upgrade Lassie from v0.4.0 to v0.5.1. Release notes: 
- https://github.com/filecoin-station/rusty-lassie/releases/tag/v0.5.0
- https://github.com/filecoin-station/rusty-lassie/releases/tag/v0.5.1

### DIFF
--- a/cli/main.rs
+++ b/cli/main.rs
@@ -41,8 +41,7 @@ async fn main_impl() -> Result<()> {
 
 #[allow(dead_code)]
 struct RunOutput {
-    // TBD
-    module_result: (),
+    module_output: (),
     // for testing
     lassie_daemon: Arc<lassie::Daemon>,
 }
@@ -91,10 +90,11 @@ async fn run_module(file: String) -> Result<RunOutput> {
         )
     };
 
-    run_js_module(&main_module, &runtime_config).await?;
+    #[allow(clippy::let_unit_value)]
+    let module_output = run_js_module(&main_module, &runtime_config).await?;
 
     Ok(RunOutput {
-        module_result: (),
+        module_output,
         lassie_daemon,
     })
 }
--- a/cli/main.rs
+++ b/cli/main.rs
@@ -130,7 +130,7 @@ mod tests {
             assert_fs::NamedTempFile::new("dummy.js").expect("cannot create temp dummy.js");
 
         mod_js
-            .write_str(&format!("/* no-op */",))
+            .write_str("/* no-op */")
             .expect("cannot write to dummy.js");
 
         let RunOutput { lassie_daemon, .. } =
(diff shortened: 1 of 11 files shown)


cli/main.rs
```diff
@@ -57,6 +58,7 @@ async fn main_impl() -> Result<()> {
                     temp_dir: None,
                     // Listen on an ephemeral port selected by the operating system
                     port: 0,
+                    access_token: Some(generate_lassie_access_token()),
```
bajtos, line 61: I tested this change manually by adding `console.log` to `runtime/js/fetch.js`, running `zinnia-dev run runtime/tests/js/ipfs_retrieval_tests.js` and observing output.

daemon/main.rs
```diff
@@ -49,6 +52,7 @@ async fn run(config: CliArgs) -> Result<()> {
         temp_dir: Some(lassie_temp_dir),
         // Listen on an ephemeral port selected by the operating system
         port: 0,
+        access_token: Some(generate_lassie_access_token()),
```
bajtos, line 55: I didn't test this change at all. It's difficult to test it since the access token is not visible to Zinnia modules.

bajtos (commented):
This change is difficult to test in an automated way 🤔

juliangruber (commented):
Could we test this by reproducing the exact case that this feature is preventing? So we get the port somehow, and try to make a direct HTTP request to lassie (doesn't matter whether that's from within Zinnia or not)

juliangruber:
Idea: Only run the test on linux, and use https://stackoverflow.com/questions/942824/how-to-find-ports-opened-by-process-id-in-linux to get the port

bajtos:
> Could we test this by reproducing the exact case that this feature is preventing? So we get the port somehow, and try to make a direct HTTP request to lassie (doesn't matter whether that's from within Zinnia or not)

See [515893a](https://github.com/filecoin-station/zinnia/pull/273/commits/515893a1f091ed0c72d5f905234c746b1375d82a)

bajtos:
> Idea: Only run the test on linux, and use https://stackoverflow.com/questions/942824/how-to-find-ports-opened-by-process-id-in-linux to get the port

That would work too 👍🏻

cli/main.rs
```diff
@@ -96,3 +116,58 @@ fn exit_with_error(error: Error) {
...
+            .expect("cannot write HTTP request");
+
+        let status = BufReader::new(stream)
+            .lines()
+            .next_line()
+            .await
+            .expect("cannot read the first line of the HTTP response")
+            .expect("server returned at least one line");
+
+        assert_eq!(status, "HTTP/1.1 401 Unauthorized")
```
juliangruber, line 171: nice!

juliangruber (approved):
This is great!

PR #354, opened 2023-08-30
deps: upgrade Deno to 1.36.3 / 2023.08.24

bajtos:
- bump `deno_core` to 0.204.0, the version used by other deno crates
- bump other deno crates to version 1.36.3 / 2023.08.24
  https://github.com/denoland/deno/releases/tag/v1.36.3

Notably, this fixes builds using recent rust versions (e.g. 1.72.0).

PR #355, opened 2023-08-30
chore: fix cargo-audit

bajtos:
Configure `cargo audit` to ignore the following two security advisories:
- [RUSTSEC-2022-0093 Double Public Key Signing Function Oracle Attack on `ed25519-dalek`](https://rustsec.org/advisories/RUSTSEC-2022-0093)
  → We are not using this cipher AFAIK
- [RUSTSEC-2023-0052 webpki: CPU denial of service in certificate path building](https://rustsec.org/advisories/RUSTSEC-2023-0052)
  → There is nothing we can do to fix this; we have to wait for Deno & libp2p to solve this.

Tweak GH Action workflow to re-run Security audit for audit config changes too.


bajtos:
> this branch seems to be based on the other feature branch you just created

yes, it is. we cannot compile the project before we upgrade Deno.

I'll clean up the commits after #354 lands.

bajtos:
@juliangruber would you like to review these changes, or should I go ahead and merge the PR?

PR #365, opened 2023-09-05
deps: upgrade deno_std to v0.201.0

bajtos:
Bring in new `assert` features, e.g. https://github.com/denoland/deno_std/pull/3496 and https://github.com/denoland/deno_std/pull/3577

### DIFF
--- a/runtime/vendor.js
+++ b/runtime/vendor.js
@@ -5,7 +5,7 @@
 // Run this script using the following command:
 //   deno run --allow-run runtime/vendor.js
 
-const STD_VERSION = "0.183.0";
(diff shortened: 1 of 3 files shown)


runtime/vendor.js
```diff
@@ -34,9 +34,8 @@ async function vendor(url, outfile) {
 async function patchAssertsBundle(assertsPath) {
   return patchFile(assertsPath, (content) =>
     content
-      .replace("const { Deno  } = globalThis;\n", "")
```
bajtos, line 37: There is more than one place using the global Deno object. It's simpler to keep it.

PR #470, opened 2024-01-24
fix: Remove full stop from started event

juliangruber:


daemon/main.rs
```diff
@@ -56,7 +56,7 @@ async fn run(config: CliArgs) -> Result<RunOutput> {
             .context("cannot initialize the IPFS retrieval client Lassie")?,
     );
 
-    log_info_activity("Module Runtime started.");
+    log_info_activity("Module Runtime started");
```
bajtos, line 59:
```suggestion
    log_info_activity("Zinnia started");
```

bajtos (approved):
While we are making this change and going through the pain of the release process, should we change the message to match what we want to display in Station, so that we no longer need to rewrite the message?

daemon/main.rs
juliangruber: Great idea!

PR #475, opened 2024-02-01
feat: `activity:started` event

bajtos:
Add a new activity event to report when the runtime started. We can expose
this new event in the JS API for Station Modules in the future too.

Example output:

```
{"type":"activity:started","module":null}
{"type":"jobs-completed","total":5,"modules":{"main":5}}
```

See https://github.com/filecoin-station/core/pull/327#discussion_r1474361506


juliangruber (approved):
Just flagging that this is a breaking change. We're still on 0.x, so 0.17.0 should be the right version

PR #526, opened 2024-04-22
feat: new Zinnia API deploymentType

PatrickNercessian:
Links:
https://github.com/filecoin-station/zinnia/pull/526

### DIFF
--- a/daemon/args.rs
+++ b/daemon/args.rs
@@ -13,6 +13,10 @@ pub struct CliArgs {
     #[arg(long, env = "STATION_ID", name = "STATION ID")]
     pub station_id: String,
 
+    /// The type of Station Core deployment. Either 'cli', 'docker', or 'station-desktop' (required).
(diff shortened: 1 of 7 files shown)


juliangruber (commented):
Why is this forwarded to the module? I thought it's Station's job to submit this data

juliangruber:
Yes I think so. Let's sync on this overall: How is data submitted to the platform?

bajtos:
Since the deployment type will not change over time, we should not record it in every measurement - that would be very wasteful.

Instead, I'd like to record information about the station instance (instance type, RAM size, etc.) using a different mechanism and then find a way how to link measurements to station instance information using the `station_id` field.

bajtos:
> Instead, I'd like to record information about the station instance (instance type, RAM size, etc.) using a different mechanism and then find a way how to link measurements to station instance information using the `station_id` field.

E.g. using an approach similar to https://github.com/filecoin-station/core/pull/411

PR #520, opened 2024-04-12
feat: new Zinnia API `stationId`

PatrickNercessian:
DRAFT: DO NOT MERGE

Links:
filecoin-station/roadmap#96

PR #533, opened 2024-05-02
feat: add regex assertion for station_id

PatrickNercessian:
Related to #520

### DIFF
--- a/daemon/main.rs
+++ b/daemon/main.rs
@@ -88,7 +88,10 @@ async fn run(config: CliArgs) -> Result<RunOutput> {
         rng_seed: None,
     };
 
-    if !Regex::new(r"^[0-9a-fA-F]{88}$").unwrap().is_match(&runtime_config.station_id) {
+    if !Regex::new(r"^[0-9a-fA-F]{88}$")
+        .unwrap()
+        .is_match(&runtime_config.station_id)
+    {
         return Err(anyhow!("Invalid station_id format"));
     }
 
--- a/daemon/main.rs
+++ b/daemon/main.rs
@@ -88,13 +88,6 @@ async fn run(config: CliArgs) -> Result<RunOutput> {
         rng_seed: None,
     };
 
-    if !Regex::new(r"^[0-9a-fA-F]{88}$")
-        .unwrap()
-        .is_match(&runtime_config.station_id)
-    {
-        return Err(anyhow!("Invalid station_id format"));
-    }
-
     // TODO: handle module exit and restart it
     // https://github.com/filecoin-station/zinnia/issues/146
     log::info!("Starting module {main_module}");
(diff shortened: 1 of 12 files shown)


PatrickNercessian:
Have not manually tested this change yet, and our automated tests don't seem to cover this sort of constraint checking (although we can start with this PR if we want).

daemon/main.rs
```diff
@@ -87,6 +88,10 @@ async fn run(config: CliArgs) -> Result<RunOutput> {
         rng_seed: None,
     };
 
+    if !Regex::new(r"^[0-9a-fA-F]{88}$").unwrap().is_match(&runtime_config.station_id) {
+        return Err(anyhow!("Invalid station_id format"));
+    }
+
```
bajtos, line 94: Let's move the validation to the top of the function, please. We already have some validation code on lines 36-43.

bajtos:
> Have not manually tested this change yet, and our automated tests don't seem to cover this sort of constraint checking (although we can start with this PR if we want).

I am fine with testing this change manually.

Please run `cargo fmt` to fix the formatting of your new code.

daemon/main.rs
PatrickNercessian: Ah I meant to leave a comment for this.

My reasoning was that if we had it with the other validations, we are not testing our own default value in the BootstrapOptions constructor.

Not sure if that has any value, what do you think?

bajtos: Ah, that's an interesting point. 

I agree it's useful to double-check the default value, too. But I don't think that's happening right now, because the default value is not used by `zinniad`.

Let's move the assertion to the function `run_js_module`, that's the entry point used by both `zinnia` and `zinniad`.

runtime/runtime.rs
```diff
@@ -97,6 +97,13 @@ pub async fn run_js_module(
     module_specifier: &ModuleSpecifier,
     bootstrap_options: &BootstrapOptions,
 ) -> Result<(), AnyError> {
+    if !Regex::new(r"^[0-9a-fA-F]{88}$")
+        .unwrap()
```
bajtos, line 103: Let's move the Regex initialisation/compilation outside of the `run_js_module` function. While this is not a hot code path, I'd like to avoid preliminary pessimization.

See https://docs.rs/regex/latest/regex/#avoid-re-compiling-regexes-especially-in-a-loop

```rs
use {
    once_cell::sync::Lazy,
    regex::Regex,
};

fn some_helper_function(haystack: &str) -> bool {
    static RE: Lazy<Regex> = Lazy::new(|| Regex::new(r"...").unwrap());
    RE.is_match(haystack)
}
```

juliangruber (approved):
Code changes Lgtm, tests would be nice but not necessary

bajtos (approved):
LGTM now 👍🏻 

I agree that tests would be nice, but I am ok to land this PR without them.

PR #539, opened 2024-05-13
ci: pin Rust version to 1.77

bajtos:
Our release binaries are crashing when built by Rust 1.78.

Suspected cause: Rust 1.78 upgraded LLVM from v17 to v18, perhaps our ancient Deno version is not compatible with that.


bajtos:
Building the release binaries here: https://github.com/filecoin-station/zinnia/actions/runs/9065519107

Issue #485, opened 2024-02-07
"Cannot import files outside of module root directory"


juliangruber:
**WORKAROUND TO FIX THE ISSUE MANUALLY**

1. Quit Station
2. Remove the whole spark folder (`AppData\Local\Temp\Filecoin Station Desktop\sources\spark`)
3. Start Station again

---

This error was submitted by a user in https://filecoinproject.slack.com/archives/C03S6LXSRB8/p1707311859950409:

```
[2024-02-07T13:25:54Z ERROR zinniad] Cannot import files outside of module root directory \\?\C:\Users\Administrator\AppData\Local\Temp\Filecoin Station Desktop\sources\voyager. file:///C:/Users/ADMINI~1/AppData/Local/Temp/Filecoin%20Station%20Desktop/sources/voyager/main.js

{"type":"activity:info","module":"Zinnia","message":"Voyager exited with code: 1"}
```

This is the first time I've seen that error, other installations are looking fine.

Some suspicions:

- Could this be a filesystem case sensitive problem?
- Is this another error masking as this one?

Zinnia source of error: https://github.com/filecoin-station/zinnia/blob/ca764ff3e919e79c46838e1d8d1936b6a944fc68/runtime/module_loader.rs#L99

Mentioned Voyager file: https://github.com/filecoin-station/voyager/blob/4f4410c36df95774e6f52cabf717c322e27e6693/main.js

bajtos:
> Could be! But I don't think so necessarily https://stackoverflow.com/questions/21194530/what-does-mean-when-prepended-to-a-file-path

Great find!

Another interesting fact:

The module toot directory contains `Administrator` among the directory names, while the module URL contains `AMINI~1`.

I suspect Deno uses a process for converting paths to strings on Windows that is different from Zinnia's.

@juliangruber What is the impact of this issue? How many users are affected?

bajtos:
I think this may be fixed by https://github.com/filecoin-station/core/pull/345. Let's see if the problem persists after people update their Station versions to apply that patch.

juliangruber:
yeah! I'm releasing the changes atm, to validate this

juliangruber:
Released in https://github.com/filecoin-station/desktop/releases/tag/v1.3.8, asked user for confirmation

juliangruber:
Let's reopen if the issue comes back

juliangruber:
I cannot reproduce this on my Windows 11 Pro laptop

[...]

PR #557, opened 2024-07-03
feat: more info when module is outside of the root

bajtos:
Let's try to get more visibility into the following issue:
- https://github.com/filecoin-station/zinnia/issues/485


juliangruber (approved):
great to get closer to an answer here 💪

PR #562, opened 2024-07-09
fix: object diff when assertion failed

bajtos:
Expose `Deno.inspect` as `Zinnia.inspect`.

Fix the error message printed by assertion functions to describe object properties.

Before this fix:

```
error: Uncaught AssertionError: Values are not equal.

    [Diff] Actual / Expected

    "[object Object]"

    at assertEquals (ext:zinnia_runtime/vendored/asserts.bundle.js:479:11)
```

After the fix:

```
error: Uncaught AssertionError: Values are not equal.

    [Diff] Actual / Expected

    {
-     foo: "bar",
+     foo: "123",
    }

    at assertEquals (ext:zinnia_runtime/vendored/asserts.bundle.js:495:11)
```

I noticed the problem while working on https://github.com/filecoin-station/spark/issues/46

### DIFF
--- a/docs/building-modules.md
+++ b/docs/building-modules.md
@@ -389,6 +389,12 @@ const response = await fetch(requestUrl, {
 
 ### Miscelaneous APIs
 
+#### `Zinnia.inspect`
+
+Converts the input into a string that has the same format as printed by `console.log()`.
+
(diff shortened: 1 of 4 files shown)


juliangruber (approved):
nice!

PR #593, opened 2024-08-27
fix: report job count only after it was persisted

bajtos:
Fixes https://github.com/filecoin-station/core/issues/583 (hopefully).

This change has a side effect when running a high-job-frequency module like the former Saturn L2 node or Voyager:
- Before, we would update the state file every time a job was completed.
- Now, we update the state file at most once per 200ms, as configured via `job_report_delay`:

    https://github.com/filecoin-station/zinnia/blob/9d882e48e5b9f0fdab2530fdc08c69987a4237a6/daemon/main.rs#L78-L82

I am arguing this is a change for the better because there is no point in hammering the file system (and SSD drives) with too frequent updates of the state file.

### DIFF
--- a/daemon/station_reporter.rs
+++ b/daemon/station_reporter.rs
@@ -60,6 +60,24 @@ impl StationReporter {
 
         print_event(&event);
     }
+
+    fn update_jobs_completed(&self, total: u64) {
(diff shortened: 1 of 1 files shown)


bajtos:
@juliangruber ping, do you have any opinion on this change? If not, then I propose to proceed to land & publish it.

PR #677, opened 2025-03-07
ci: upgrade Rust to v1.85 + maintenance

bajtos:
- **test: retrieve the CID hosted by our Frisbii**
- **fix: remove unused struct field**
- **ci: bump Rust to v1.85**

Close https://github.com/CheckerNetwork/zinnia/issues/676

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -12,6 +12,7 @@ concurrency:
 
 env:
   SEGMENT_DOWNLOAD_TIMEOUT_MINS: 2 # Fail cache download after 2 minutes.
+  RUST_VERSION: '1.85'
 
 jobs:
   build:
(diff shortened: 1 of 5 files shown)


bajtos:
I am very surprised that upgrading Rust from v1.77 to v1.85 did not break anything inside the Deno/V8 integration. 

We have a large test suite executing JavaScript files to verify that all APIs work (see [runtime/tests/js](https://github.com/CheckerNetwork/zinnia/tree/main/runtime/tests/js)), so if the CI passes on all platforms, then this upgrade should be safe to perform.

ext/libp2p/peer.rs
```diff
@@ -280,7 +280,6 @@ impl EventLoop {
                 }
 
                 RequestResponseEvent::Message {
-                    peer: _,
```
pyropy, line 283: Great that you found this piece of dead code 👏🏻

.github/workflows/ci.yaml
```diff
@@ -12,6 +12,7 @@ concurrency:
 
 env:
   SEGMENT_DOWNLOAD_TIMEOUT_MINS: 2 # Fail cache download after 2 minutes.
+  RUST_VERSION: '1.85'
```
pyropy, line 15: Could you also do the same for go version? 🙏🏻  Or should we open separate PR for that?

pyropy (approved):
Great work Miro! I'm going approve it since the PR is ready to be shipped as is. I have left one comment that can be addressed in separate PR or it can be included in this one too (it's up to you)

.github/workflows/ci.yaml
bajtos: Let's open a separate PR for that please.

PR #690, opened 2025-03-14
fix: Linux builds

pyropy:
Fixes regression introduced by #668 by upgrading golang toolchain version from `v1.20` to `v1.22` as expected by `rusty-lassie`. 

Related to #691 
Partially closes #689

PR #691, opened 2025-03-14
fix: Downgrade rust toolchain version to v1.77

### LINKED ISSUE #689: Fix Zinnia release builds
**Linux**

https://github.com/CheckerNetwork/zinnia/actions/runs/13852124103/job/38761121022

```
  Building /target/x86_64-unknown-linux-gnu/release/build/lassie-247eac7cef2dfbbc/out/libgolassie.a for x86_64 (GOARCH=amd64)
  go: errors parsing go.mod:
  /cargo/registry/src/index.crates.io-1949cf8c6b5b557f/lassie-0.10.0/go.mod:3: invalid go version '1.22.0': must match format 1.23
```

(I think we need to find a way to configure the Docker image used by `cross` to install a different Go version.)

**macOS**

```
❯ ~/Downloads/zinnia run manual-check.js
Calling Filecoin JSON-RPC to get PeerId of miner f03160660
zinnia(20960,0x204360840) malloc: *** error for object 0x102a4462c: pointer being freed was not allocated
zinnia(20960,0x204360840) malloc: *** set a breakpoint in malloc_error_break to debug
```

(I suspect our ancient Deno version is incompatible with the latest Rust version.)

**Windows**

After fixing all other issues, we must verify that the new release binary can perform Spark manual check when using the Graphsync protocol.

pyropy:
Fixes regression introduced by #677 and #688 by downgrading rust toolchain version from `v1.85` to `v1.77`

### Changelog
- Downgrades rust toolchain version to `1.77`
- Downgrades Cargo.lock from version `4` to version `3`

Related to #690 
Closes #689


Cargo.lock
```diff
@@ -1,6 +1,6 @@
 # This file is automatically @generated by Cargo.
 # It is not intended for manual editing.
-version = 4
+version = 3
```
bajtos, line 3: Isn't it enough to change the version of the Cargo.lock file and keep the dependencies at their latest versions?

(I.e. keep the change above and revert the rest.)

pyropy: You're right. I was hesitant to do this by hand so I had it deleted and tried to generate a new file, but I was getting errors when trying to compile some dependencies. 

I have reverted Cargo files back to version from main, downgraded Cargo.lock version to `3` manually and had the runtime package version bumped in commit https://github.com/CheckerNetwork/zinnia/pull/691/commits/b1458e344bc9f69bbfc84f1f531f7442f884b59a

Cargo.lock
```diff
@@ -4934,7 +4934,7 @@ dependencies = [
 
 [[package]]
 name = "zinnia"
-version = "0.21.1"
+version = "0.22.0"
```
bajtos, line 4937: I think you should revert all of these version changes for local packages in this monorepo. They will conflict with the changes made on the `main` branch.

Cargo.toml
```diff
@@ -27,8 +27,8 @@ serde_repr = "0.1.20"
 tokio = { version = "1.44.1", features = ["fs", "rt", "macros"] }
 
 # workspace-local
-zinnia_runtime = { version = "0.21.1", path = "./runtime" }
-zinnia_libp2p = { version = "0.21.1", path = "./ext/libp2p" }
+zinnia_runtime = { version = "0.22.0", path = "./runtime" }
+zinnia_libp2p = { version = "0.22.0", path = "./ext/libp2p" }
```
bajtos, line 31: Ditto. I think you should revert all of these version changes for local packages in this monorepo. They will conflict with the changes made on the `main` branch.

Same comment applies to other `Cargo.toml` in this repository.

bajtos (commented):
Please merge the latest `main` to run the tests in release build for your fix.

Cargo.lock
pyropy: Resolved in https://github.com/CheckerNetwork/zinnia/pull/691/commits/4d4d00e6c3b605b523d1d902b03a5700a3519f4e

PR #692, opened 2025-03-14
ci: test release builds for all changes

bajtos:
Prevent regressions like https://github.com/CheckerNetwork/zinnia/issues/689 from happening again.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -50,17 +50,17 @@ jobs:
           # commits added to an already-opened pull request.
           # save-if: ${{ github.ref == 'refs/heads/main' }}
 
-
       - run: cargo build --workspace --all-features --all-targets --target=${{ matrix.target }}
 
   test:
-    name: Test on ${{ matrix.target }}
+    name: Test on ${{ matrix.target }} ${{ matrix.release-mode && '(release)' || '(debug)' }}
     runs-on: ${{ matrix.os }}
     needs:
       - build
     strategy:
       fail-fast: false
       matrix:
+        release-mode: [false, true]
         include:
           - target: "x86_64-apple-darwin"
             os: macos-latest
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yaml
```diff
@@ -17,19 +17,19 @@ env:
...
 jobs:
   build:
-    name: Build on ${{ matrix.target }}
+    name: Build on ${{ matrix.target.triple }}
     strategy:
       matrix:
-        include:
-          - target: x86_64-apple-darwin
+        target:
+          - triple: x86_64-apple-darwin
```
pyropy, line 24: Should we maybe rename `triple` to `arch` or something like that? I'm not exactly sure what `triple` should represent.

pyropy (changes requested):

.github/workflows/ci.yaml
bajtos: "Target triple" is a term used by Rust - see https://doc.rust-lang.org/nightly/rustc/platform-support.html

> Targets are identified by their "target triple" which is the string to inform the compiler what kind of output should be produced.

bajtos:
Yay, the release builds failed, so we have a reproduction of the problem 👍🏻 

https://github.com/CheckerNetwork/zinnia/actions/runs/13859131946/job/38783135647#step:6:410

```
running 1 test
memory allocation of 140267894385153 bytes failed
error: test failed, to rerun pass `-p zinnia_runtime --test fetch_api_tests`
```

That one is funky - allocating 140267 gigabytes of memory???

.github/workflows/ci.yaml
```diff
@@ -50,17 +50,17 @@ jobs:
...
   test:
-    name: Test on ${{ matrix.target }}
+    name: Test on ${{ matrix.target }} ${{ matrix.release-mode && '(release)' || '(debug)' }}
     runs-on: ${{ matrix.os }}
     needs:
       - build
     strategy:
       fail-fast: false
       matrix:
+        release-mode: [false, true]
```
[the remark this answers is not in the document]
bajtos: Fixed.

.github/workflows/ci.yaml
pyropy: In that case I think we're okay with using the term `triple`. Thanks for providing the resource!

pyropy (approved):
Good job on reproducing the issue 👏🏻

PR #715, opened 2025-04-07
ci: run CI on push to any branch

bajtos:
I noticed that GitHub is not running the `ci.yaml` workflow for new commits pushed to the `main` branch. In this PR, I am removing the branch restriction from the trigger config.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -3,8 +3,6 @@ name: Continuous integration
 on:
   pull_request:
   push:
-    branches:
-      - main
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}


bajtos:
Also, we used to run CI on the main branch in the past, see https://github.com/CheckerNetwork/zinnia/actions/workflows/ci.yaml?query=branch%3Amain

Screenshot 2025-04-07 at 11 11 08

- The last commit for which we ran CI on main: 65ee57ac
- The first commit where CI was not triggered: 3f8ae368

bajtos:
Well, after landing this PR, new commits pushed to `main` are triggering the CI build again. I still don't understand why 🤷🏻, but it's great that the problem is fixed.

Screenshot 2025-04-07 at 11 15 13

juliangruber:
@bajtos I think something else is at play. It's the situation we're familiar with: Dependabot commits not triggering CI. If you look at recent commit, all created by humans had CI run on them

bajtos:
> @bajtos I think something else is at play. It's the situation we're familiar with: Dependabot commits not triggering CI. If you look at recent commit, all created by humans had CI run on them

Great insight about dependabot commits not triggering the CI builds.

Do you remember how we solved that problem in other repositories?

I checked spark-api and could not find anything special in `.github/workflows` or repository settings.

When I look at the CI workflow in this repo, there is one step using GitHub secrets - do you think that is preventing GitHub from running the job on the `main` branch?

https://github.com/CheckerNetwork/zinnia/blob/11c1c0d84816f4e28fdcc151a63036059983e451/.github/workflows/ci.yaml#L178-L197

PR #717, opened 2025-04-09
feat: upgrade Deno to v2.2.8 / 2025.04.05

bajtos:
Notable changes:

- Upgrade Rust to 1.85
- Zinnia can run TypeScript files now
- Global `window` was removed (see https://github.com/denoland/deno/pull/22057)
- Fetch API: `Request.bytes()` and `Response.bytes()`
- WebCrypto API: import and export p521 keys, X448 support
- Web API: `Blob.prototype.bytes()`
- Import attributes & importing JSON files
- `URL.parse()`
- `ReadableStream.from()` allows `Iterable` instead of `IterableIterator`
- `Float16Array`
- WASM module support
- Fix `AbortSignal.timeout()` leak
- Fix potential leak of unread buffers
- Prevent (De-)CompressionStream resource leak on stream cancellation

TODO:

- [ ] fix stderr output for uncaught errors
- [ ] test that `fetch://` rejects `file://` urls
- [ ] test importing JSON files
- [ ] test running TS files + error stack traces (do we need sourcemaps?)
- [ ] test `URL.parse()`
- [ ] test `Float16Array`
- [ ] test `import.meta.filename` and `import.meta.dirname`
- [ ] test `Request.bytes()` and `Response.bytes()`
- [ ] test WAM module support

Links:
- https://gist.github.com/bajtos/9cd1cf2dc9d22227eef68a3a356541ff


cli/tests/cli_tests.rs
```diff
@@ -70,11 +70,10 @@ function fail() {
         &std::env::current_dir().context("unable to get current working directory")?,
     )?;
 
+    // FIXME: stderr should include the source code line constructing the error
     let expected_stderr = format!(
         r#"
-  error: Uncaught Error: boom!
-  throw new Error("boom!");
-        ^
+  error: Uncaught (in promise) Error: boom!
```
bajtos, line 76: This is a regression I'd like to fix, but it's not blocking this PR from landing.

runtime/js/01_version.ts
bajtos, line 1: Notice the file has `.ts` extension. Deno started to use TS for internal files, including files in the extensions we are using. While I could use `.js` extension here, I decided to use `.ts` for consistency with `01_version.ts` in `deno_runtime`.

bajtos:
@NikolasHaimerl @juliangruber @pyropy 

This pull request is ready for review and landing.

I don't expect you to understand all the details - you would have to dig into Deno's codebase as I did. Skim through the patch and look for any obvious problems or suspicious changes.

Depending on how quickly we can get this approved, I will either add a small commit to fix the regression in error formatting to this pull request or open a follow-up pull request to fix that. I consider this regression as a minor issue not blocking this PR from being landed.

There are no tests for the new features introduced by Deno and listed in the PR description. I'll open follow-up pull requests for that; there is a checklist at https://github.com/CheckerNetwork/zinnia/issues/676.

runtime/js/98_global_scope.js
```diff
@@ -17,118 +16,118 @@ import * as url from "ext:deno_url/00_url.js";
 import * as urlPattern from "ext:deno_url/01_urlpattern.js";
 import * as headers from "ext:deno_fetch/20_headers.js";
 import * as streams from "ext:deno_web/06_streams.js";
-// Unused import, required to work around Deno's check:
-// "Following modules were not evaluated; make sure they are imported from other code"
-import * as fileReader from "ext:deno_web/10_filereader.js";
+// import * as fileReader from "ext:deno_web/10_filereader.js";
```
pyropy, line 19: Are these unusded imports still required as a workaround? If so let's add a comment above them like we use to have (see the diff).

runtime/ext.rs
```diff
@@ -27,11 +30,66 @@ impl TimersPermission for ZinniaPermissions {
...
+        _resolved: bool,
+        _p: &'a Path,
+        _api_name: &str,
+    ) -> Result<Cow<'a, Path>, FsError> {
+        Err(FsError::NotCapable("read local filesystem"))
+    }
+}
+
+impl NetPermissions for ZinniaPermissions {
+    fn check_net<T: AsRef<str>>(
```
pyropy, line 47: Will the user running checker app and or zinnia runtime have to grant these permissions to the app?

bajtos:
> Could we also update Zinnia requirements in the [cli/README.md](https://github.com/CheckerNetwork/zinnia/blob/main/cli/README.md#build-from-source) before we merge? 🙏🏻

Great catch. I updated the instructions in c698365

runtime/ext.rs
bajtos: No, the Deno Net API operations are not permitted in Zinnia. 

At the moment, Zinnia modules are allowed to use the `fetch()` API only, which uses a different permission check.

In the future, if we decide to expose Deno Net APIs, then we can change the implementation of this permission check.

runtime/js/98_global_scope.js
pyropy: Oh okay, got it!

pyropy (approved):
Great work Miro! 🚀

cli/tests/cli_tests.rs
juliangruber: Did you open a follow up tech debt issue?

cli/Cargo.toml
```diff
@@ -32,9 +32,9 @@ tokio = { workspace = true, features = ["io-util"] }
 [package.metadata.winres]
 # This section defines the metadata that appears in the deno.exe PE header.
 OriginalFilename = "zinnia.exe"
-LegalCopyright = "© Protocol Labs & Zinnia contributors. Permissive License Stack (Apache-2.0 or MIT)."
-ProductName = "Filecoin Station"
-FileDescription = "Zinnia: Secure sandboxed runtime for Filecoin Station modules."
+LegalCopyright = "© Space Meridian & Zinnia contributors. Permissive License Stack (Apache-2.0 or MIT)."
+ProductName = "Checker Network"
+FileDescription = "Zinnia: Secure sandboxed runtime for Checker Network checker nodes"
```
juliangruber, line 37: Let's call these Checker Network nodes, Checker Network checker nodes is redundant and doesn't sound good to me

cli/tests/cli_tests.rs
juliangruber: looks like this was fixed in https://github.com/CheckerNetwork/zinnia/pull/717/commits/1874fc8dc173291de6ecd113b229070911ff84e0 and this conversation can be resolved

juliangruber (approved):
Approving with minor change requests

Massive effort, @bajtos!!


it almost feels impossible

runtime/ext.rs
```diff
@@ -96,3 +157,24 @@ fn op_zinnia_log(state: &mut OpState, #[string] msg: &str, #[smi] level: i32) {
 fn op_format_test_error(#[serde] error: JsError) -> String {
     crate::vendored::cli_tools::format_test_error(&error)
 }
+
+// https://github.com/denoland/deno/blob/v2.2.8/runtime/ops/bootstrap.rs
+// Copyright 2018-2025 the Deno authors. MIT license.
+
```
NikolasHaimerl, line 163:
```suggestion
```

runtime/js/98_global_scope.js
```diff
@@ -17,118 +16,118 @@ import * as url from "ext:deno_url/00_url.js";
...
+  ReadableByteStreamController: core.propNonEnumerable(streams.ReadableByteStreamController),
+  ReadableStreamBYOBReader: core.propNonEnumerable(streams.ReadableStreamBYOBReader),
+  ReadableStreamBYOBRequest: core.propNonEnumerable(streams.ReadableStreamBYOBRequest),
+  ReadableStreamDefaultController: core.propNonEnumerable(streams.ReadableStreamDefaultController),
+  TransformStreamDefaultController: core.propNonEnumerable(
+    streams.TransformStreamDefaultController,
+  ),
+  atob: core.propWritable(base64.atob),
+  btoa: core.propWritable(base64.btoa),
+  // TODO: ImageData & ImageBitmap
```
NikolasHaimerl, line 98: Should we create an link an issue here?

NikolasHaimerl (approved):

runtime/ext.rs
```diff
@@ -42,12 +100,15 @@ deno_core::extension!(
         op_info_activity,
         op_error_activity,
         op_zinnia_log,
-        op_format_test_error
+        op_format_test_error,
+
```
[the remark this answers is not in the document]
bajtos: I am using the empty line to delimit `op_bootstrap_*` ops from the rest of the pack. In Deno, these ops were extracted into a new extension. I feel it would be over-engineering to add a new extension in Zinnia, so I am separating the bootstrapping ops visually.

runtime/ext.rs
bajtos: The comment above applies to all code below, not only the next function. That's why I have an empty line.

Now that I am thinking about this - would it be better to move `op_bootstrap_*` functions to a new file?

cli/Cargo.toml
bajtos: I share the same concerns.

FWIW, I used the term "checker nodes" because we have other kinds of nodes too - the nodes in the backend/orchestration layer.

I am going to use `runtime for Checker Network nodes`, we can easily change this later as needed.

bajtos: See af0030e

runtime/js/98_global_scope.js
bajtos: Great suggestion! See d3412ce

PR #721, opened 2025-04-09
ci: run macos tests on aarch64

bajtos:
macos-latest (macos-14) is running on Apple Silicon. When we cross-compile to `x86_64`, the build takes longer to complet and is failing in #717.

I don't have the apetite to dig deeper to find the root cause, changing the runner images used by build and test steps is the easiest fix.


bajtos:
The tests are failing because they cannot download pre-build v8. I suspect our version is so old that there is no other way around that and we have to land the upgrade first (see #717)

PR #722, opened 2025-04-11
feat: load WASM modules

bajtos:
Example use:

```js
const { add } = await import("math.wasm");
const result = add(2, 3);
assertEquals(result, 5);
```

Links:
- https://deno.com/blog/intro-to-wasm
- https://github.com/CheckerNetwork/zinnia/issues/706


juliangruber (commented):
Would you label this as full WASM support, or is it basic WASM support? I'm trying to understand the complexities / requirements surrounding loading WASM modules

bajtos:
> Would you label this as full WASM support, or is it basic WASM support? I'm trying to understand the complexities / requirements surrounding loading WASM modules

This is not full support (most likely).

It's a small improvement over what we have in the "old" zinnia, where you had to include WASM bytes in your JS code (IIRC, see https://github.com/CheckerNetwork/runtime-poc/tree/main/deno-core). 

My goal here is to close the gap introduced by https://github.com/CheckerNetwork/zinnia/pull/717, where we upgraded to the Deno version that can import `.wasm` files directly but where we did not implement the necessary wiring on the Zinnia side.

As the next step for [PoC for Zinnia modules built using Rust/WASM](https://github.com/CheckerNetwork/zinnia/issues/706), I want to create a small example module with Rust code calling out to Web & Zinnia APIs. The example will document how to do it and identify the pain points.

juliangruber (approved):

NikolasHaimerl (approved):

PR #729, opened 2025-04-14
test: avoid IPNI lookups

bajtos:
IPNI service is often downgraded and our tests fail as a result.

This commit adds `protocol` and `providers` url query parameters to all Lassie requests so that Lassie can skip the provider lookup.

A similar change for rusty-lassie:
- https://github.com/CheckerNetwork/rusty-lassie/pull/164

### DIFF
--- a/runtime/tests/js/ipfs_retrieval_tests.js
+++ b/runtime/tests/js/ipfs_retrieval_tests.js
@@ -2,13 +2,13 @@ import { test } from "zinnia:test";
 import { assertEquals, assertMatch, assertRejects, AssertionError } from "zinnia:assert";
 
 const TEST_CID = "bafkreih25dih6ug3xtj73vswccw423b56ilrwmnos4cbwhrceudopdp5sq";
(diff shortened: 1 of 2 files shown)


runtime/tests/js/ipfs_retrieval_tests.js
```diff
@@ -2,12 +2,13 @@ import { test } from "zinnia:test";
 import { assertEquals, assertMatch, assertRejects, AssertionError } from "zinnia:assert";
 
 const TEST_CID = "bafkreih25dih6ug3xtj73vswccw423b56ilrwmnos4cbwhrceudopdp5sq";
+const RETRIEVE_FROM_OUR_FRISBII = "?protocol=http&providers=/dns4/frisbii.fly.dev/https";
```
pyropy, line 5: I'm having issues with this name. I propose changing it to something like `SPACE_MERIDIAN_FRISBII_INSTANCE` or `FRISBII_INSTANCE_QUERY_PARAMS`?

pyropy (commented):
Awesome work Miro! 🙌🏻

pyropy (approved):

PR #735, opened 2025-04-15
test: import.meta.{filename,dirname}

bajtos:
This is a follow-up for https://github.com/CheckerNetwork/zinnia/pull/717 where we upgraded Deno to the version that supports `import.meta`.

Parent issue:


NikolasHaimerl (approved):
LGTM. Could I ask you to add a comment to each test describing what we are testing and why? 
Without the issue description the tests are not quite self explainatory IMHO.

PR #736, opened 2025-04-15
fix: source code paths relative to module root

bajtos:
We must not leak the filesystem structure when the code is running inside a sandboxed environment. Only path relative to the project root should be available to the module.

I discovered this issue while working on https://github.com/CheckerNetwork/zinnia/pull/735

Parent issue:

### DIFF
--- a/runtime/module_loader.rs
+++ b/runtime/module_loader.rs
@@ -108,6 +108,8 @@ impl ModuleLoader for ZinniaModuleLoader {
                 ModuleLoaderError::from(JsErrorBox::generic(msg))
             })?;
 
+            let sandboxed_path: PathBuf;
+
             // Check that the module path is inside the module root directory
             if let Some(canonical_root) = &module_root {
                 // Resolve any symlinks inside the path to prevent modules from escaping our sandbox
@@ -120,19 +122,23 @@ impl ModuleLoader for ZinniaModuleLoader {
                     ModuleLoaderError::from(JsErrorBox::generic(msg))
                 })?;
 
-                if !canonical_module.starts_with(canonical_root) {
-                    let msg = format!(
-                        "Cannot import files outside of the module root directory.\n\
+                let relative_path =
+                    canonical_module.strip_prefix(canonical_root).map_err(|_| {
+                        let msg = format!(
+                            "Cannot import files outside of the module root directory.\n\
                          Root directory (canonical): {}\n\
                          Module file path (canonical): {}\
                          {}",
-                        canonical_root.display(),
-                        canonical_module.display(),
-                        details()
-                    );
-
-                    return Err(ModuleLoaderError::from(JsErrorBox::generic(msg)));
-                }
+                            canonical_root.display(),
+                            canonical_module.display(),
+                            details()
+                        );
+                        ModuleLoaderError::from(JsErrorBox::generic(msg))
+                    })?;
+                sandboxed_path = Path::new(r"/").join(relative_path).to_owned();
+                // sandboxed_path = PathBuf::from_str(r".").push(relative_path);
+            } else {
+                sandboxed_path = module_path.to_owned();
             };
 
             // Based on https://github.com/denoland/roll-your-own-javascript-runtime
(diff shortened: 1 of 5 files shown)


juliangruber (commented):
Idea sounds great to me, I'll step back in favor of Nikolas' review here

runtime/module_loader.rs
```diff
@@ -120,19 +122,28 @@ impl ModuleLoader for ZinniaModuleLoader {
                     ModuleLoaderError::from(JsErrorBox::generic(msg))
                 })?;
 
-                if !canonical_module.starts_with(canonical_root) {
-                    let msg = format!(
-                        "Cannot import files outside of the module root directory.\n\
+                let relative_path =
+                    canonical_module.strip_prefix(canonical_root).map_err(|_| {
```
NikolasHaimerl, line 126: Are we sure we want to dismiss the root error?

runtime/tests/runtime_integration_tests.rs
```diff
@@ -94,8 +94,71 @@ js_tests!(ipfs_retrieval_tests);
...
+
 // Run all tests in a single JS file
 async fn run_js_test_file(name: &str) -> Result<(Vec<String>, Option<AnyError>), AnyError> {
+    run_js_test_file_with_module_root(name, None).await
+}
+
+async fn run_js_test_file_with_module_root(
+    name: &str,
+    module_root: Option<PathBuf>,
+) -> Result<(Vec<String>, Option<AnyError>), AnyError> {
```
NikolasHaimerl, line 161: Can we trim this return type down to: 
`Result<Vec<String>,AnyError>` or` Result<Option<Vec<String>,AnyError>`?

runtime/tests/runtime_integration_tests.rs
```diff
@@ -94,8 +94,71 @@ js_tests!(ipfs_retrieval_tests);
 test_runner_tests!(passing_tests);
 test_runner_tests!(failing_tests expect_failure);
 
+#[tokio::test]
+async fn source_code_paths_when_no_module_root() -> Result<(), AnyError> {
+    let (activities, run_error) =
+        run_js_test_file_with_module_root("print_source_code_paths.js", None).await?;
+    if let Some(err) = run_error {
```
NikolasHaimerl, line 101: I find the return structure a bit odd here. Can we restructure the return types so we either unwrap the error directly or propagate it using `?` ?

runtime/tests/runtime_integration_tests.rs
```diff
@@ -94,8 +94,71 @@ js_tests!(ipfs_retrieval_tests);
...
+#[tokio::test]
+async fn source_code_paths_when_inside_module_root() -> Result<(), AnyError> {
+    let module_root = Some(PathBuf::from(env!("CARGO_MANIFEST_DIR")));
+    let (activities, run_error) =
+        run_js_test_file_with_module_root("print_source_code_paths.js", module_root).await?;
+    if let Some(err) = run_error {
+        return Err(err);
+    }
+
+    let expected = if cfg!(target_os = "windows") {
```
NikolasHaimerl, line 132: How can we be sure here, that we are testing it for both windows and `other`?

runtime/module_loader.rs
bajtos: Great question!

See
https://doc.rust-lang.org/std/path/struct.Path.html#method.strip_prefix

Screenshot 2025-04-22 at 18 10 37

1. This error should never happened, because we already checked that the path starts with the prefix.
2. The error does not convey any information, see https://doc.rust-lang.org/std/path/struct.StripPrefixError.html
   ```rs
   pub struct StripPrefixError(/* private fields */);
   ```

runtime/tests/runtime_integration_tests.rs
bajtos: Great question! Let's discuss this in https://github.com/CheckerNetwork/zinnia/pull/740#discussion_r2049002936.

I want to clean this up after the current PRs are landed.

runtime/tests/runtime_integration_tests.rs
bajtos: There is `if` branch for Windows-style paths and then the `else` branch for Unix (non-Windows)-style paths. The only difference is in what paths we expect (`/ZINIA/...` or `C:\ZINNIA\...`). We need this because paths are different on Windows. 

We run the tests on Linux, macOS and Windows. The first two platforms will execute the "on non-Windows" branch, the Windows will execute the "on Windows" branch.

Does it make sense now? Maybe I don't understand your question.

runtime/module_loader.rs
NikolasHaimerl: Thanks for the info!

runtime/tests/runtime_integration_tests.rs
NikolasHaimerl: I see, I wasn't sure whether the CI runs all three platforms, now I know! :)

NikolasHaimerl (approved):

runtime/tests/runtime_integration_tests.rs
bajtos: See https://github.com/CheckerNetwork/zinnia/pull/752

runtime/tests/runtime_integration_tests.rs
bajtos: See https://github.com/CheckerNetwork/zinnia/pull/752

PR #740, opened 2025-04-17
fix: error stack traces in transpiled TypeScript

bajtos:
This is a follow-up for #717, where we added TypeScript transpilation but did not test how it works in practice.

The implementation is based on this Deno example:
https://github.com/denoland/deno_core/blob/0.343.0/core/examples/ts_module_loader.rs

Example code:

```ts
interface User {
  name: string;
  email: string;
}

const err: Error = new Error("boom!"); throw err;
```

Deno output:

```
error: Uncaught (in promise) Error: boom!
const err: Error = new Error("boom!"); throw err;
                   ^
    at file:///project/x.ts:6:20
```

Zinnia before this change - notice the missing `: Error` snippet and wrong source code line & column numbers:

```
error: Uncaught (in promise) Error: boom!
const err = new Error("boom!");
            ^
    at file:///project/x.ts:1:13
```

Zinnia after this change (same output as from Deno):

```
error: Uncaught (in promise) Error: boom!
const err: Error = new Error("boom!"); throw err;
                   ^
    at file:///project/x.ts:6:20
```

Links:
- https://github.com/CheckerNetwork/zinnia/issues/30

Related:
- https://github.com/denoland/deno/issues/4499
- https://github.com/denoland/deno_core/pull/819

### DIFF
--- a/runtime/module_loader.rs
+++ b/runtime/module_loader.rs
@@ -1,6 +1,8 @@
+use std::borrow::Cow;
+use std::cell::RefCell;
 use std::collections::HashMap;
 use std::path::{Path, PathBuf};
-use std::sync::{Arc, RwLock};
+use std::rc::Rc;
 
 use deno_ast::{MediaType, ParseParams};
 use deno_core::anyhow::anyhow;
@@ -21,7 +23,9 @@ use deno_core::anyhow::Result;
 pub struct ZinniaModuleLoader {
     module_root: Option<PathBuf>,
     // Cache mapping file_name to source_code
-    code_cache: Arc<RwLock<HashMap<String, String>>>,
+    code_cache: Rc<RefCell<HashMap<String, String>>>,
+    // Cache mapping module_specifier string to source_map bytes
+    source_maps: Rc<RefCell<HashMap<String, Vec<u8>>>>,
 }
 
 impl ZinniaModuleLoader {
@@ -34,7 +38,8 @@ impl ZinniaModuleLoader {
 
         Ok(Self {
             module_root,
-            code_cache: Arc::new(RwLock::new(HashMap::new())),
+            code_cache: Rc::new(RefCell::new(HashMap::new())),
+            source_maps: Rc::new(RefCell::new(HashMap::new())),
         })
     }
 }
@@ -79,6 +84,7 @@ impl ModuleLoader for ZinniaModuleLoader {
         let module_root = self.module_root.clone();
         let maybe_referrer = maybe_referrer.cloned();
         let code_cache = self.code_cache.clone();
+        let source_maps = self.source_maps.clone();
         let module_load = async move {
             let spec_str = module_specifier.as_str();
 
@@ -177,6 +183,10 @@ impl ModuleLoader for ZinniaModuleLoader {
 
             let code = read_file_to_string(&module_path).await?;
 
+            code_cache
+                .borrow_mut()
+                .insert(spec_str.to_string(), code.clone());
+
             let code = if should_transpile {
                 let parsed = deno_ast::parse_module(ParseParams {
                     specifier: module_specifier.clone(),
@@ -187,7 +197,7 @@ impl ModuleLoader for ZinniaModuleLoader {
                     maybe_syntax: None,
                 })
                 .map_err(JsErrorBox::from_err)?;
-                parsed
+                let res = parsed
                     .transpile(
                         &deno_ast::TranspileOptions {
                             imports_not_used_as_values: deno_ast::ImportsNotUsedAsValues::Error,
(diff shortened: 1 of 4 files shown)


runtime/tests/runtime_integration_tests.rs
```diff
@@ -94,6 +95,36 @@ js_tests!(ipfs_retrieval_tests);
...
+"#;
+        assert_eq!(actual_error.trim(), expected_error.trim());
+    } else {
+        panic!("The script threw unexpected error: {}", error);
+    }
+    Ok(())
+}
+
 // Run all tests in a single JS file
 async fn run_js_test_file(name: &str) -> Result<(Vec<String>, Option<AnyError>), AnyError> {
```
bajtos, line 129: In retrospect, it was a mistake to use `Option<AnyError>` here. I would like to rework this to return `Result<(), AnyError>` instead, but such refactoring is out of scope of this pull request.

@NikolasHaimerl, do you have a suggestion on how to best represent a type with the following variants in Rust?

- We cannot execute the JS/TS file -> `Err(AnyError)`
- We executed the JS/TS file and collected some logs in `Vec<String>`. In addition to these logs, we want to provide:
  - The script executed cleanly -> `()` 
  - The script threw a runtime error -> `AnyError`

Variants:
1. `Err(AnyError)`
2. `Ok((Ok(()), Vec<String>))`
3. `Ok((Err(AnyError), Vec<String>))`

runtime/tests/runtime_integration_tests.rs
```diff
@@ -94,6 +95,36 @@ js_tests!(ipfs_retrieval_tests);
...
+
+    if let Some(CoreError::Js(e)) = any_and_jserrorbox_downcast_ref::<CoreError>(&error) {
+        let actual_error = format_js_error(e);
+        // Strip ANSI codes (colors, styles)
+        let actual_error = console_static_text::ansi::strip_ansi_codes(&actual_error);
+        // Replace current working directory in stack trace file paths with a fixed placeholder
+        let cwd_url = ModuleSpecifier::from_file_path(std::env::current_dir().unwrap()).unwrap();
+        let actual_error = actual_error.replace(cwd_url.as_str(), "file:///project-root");
+        // Normalize line endings to Unix style (LF only)
+        let actual_error = actual_error.replace("\r\n", "\n")
```
bajtos, line 113: This is similar to the code we have in `format_test_activities()` below, but not exactly the same. WDYT, is it worth extracting a shared helper function?

https://github.com/CheckerNetwork/zinnia/blob/8680101d4e8a72da317fea905d28e7abd7ef3a3c/runtime/tests/runtime_integration_tests.rs#L153-L189

runtime/tests/runtime_integration_tests.rs
NikolasHaimerl: How about we wrap the logs inside the[ Anyhow context string](https://docs.rs/anyhow/latest/anyhow/trait.Context.html)?

bajtos: I see what you mean. We need to extract the logs from the results, so that we can assert what exactly was logged by the test runner on failure. Do you know how to extract one piece of a context from the Anyhow error?

bajtos: I am leaning towards using `Result<(Result<(), AnyError>, Vec<String>), AnyError>`.
I.e. `Result<ScriptResult, AnyError>` where `ScriptResult` is a tuple `(Result<(), AnyError>, Vec<String>)`.

bajtos: Would it help to create a named struct `ScriptResult`?

Alternatively, maybe we can treat the case "we cannot execute the script" the same way as "the script threw an error".

That will give us `(Result<(), AnyError>, Vec<String>)`

- We cannot execute the JS/TS file -> `(Err(AnyError), vec![])`
- The script executed cleanly -> `(Ok(()), logs)`
- The script threw an error -> `(Err(AnyError), logs)`

WDYT?

I'd also like to make this refactoring as a follow-up, after I land the remaining two pull requests.

runtime/module_loader.rs
```diff
@@ -21,7 +23,9 @@ use deno_core::anyhow::Result;
 pub struct ZinniaModuleLoader {
     module_root: Option<PathBuf>,
     // Cache mapping file_name to source_code
-    code_cache: Arc<RwLock<HashMap<String, String>>>,
+    code_cache: Rc<RefCell<HashMap<String, String>>>,
```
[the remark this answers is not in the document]
bajtos: Great question! I am not sure, TBH.

We are running the Deno runtime on the current thread, see here:

https://github.com/CheckerNetwork/zinnia/blob/2d403760d6afd7b2777d3e6d807704df764fb55c/daemon/main.rs#L21

https://github.com/CheckerNetwork/zinnia/blob/2d403760d6afd7b2777d3e6d807704df764fb55c/cli/main.rs#L17

I think that means all async tasks will be executed on the current thread too, including the asynchronous work of the module loader.

Deno uses `Rc<RefCell<...>>` in their example:

https://github.com/denoland/deno_core/blob/0.343.0/core/examples/ts_module_loader.rs

My knowledge of async Rust is very limited, I don't understand this deeply enough. What would you suggest to use?

runtime/tests/runtime_integration_tests.rs
NikolasHaimerl: I see what you mean. Ideally I would like to avoid having a result inside a result. We could use enums as error types to give us specific information about which error was caused by the test runner. Why do we not unwrap an error once it occurs? What is the benefit of propagating it upwards if it is only used in testing? 
Usually, in Rust testing, I would expect Errors to be unwrapped as they occur within the testing suite. Error propagation, I would usually only do in production.

runtime/module_loader.rs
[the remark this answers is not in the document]
NikolasHaimerl: Ah I see, well `Rc<RefCell<...>>` is used in a single-threaded environment, since Deno is using that as an example, I would assume it is single threaded. If we were to use multithreaded environments, I'd rather go for a `Arc<Mutex<...>>` type.

runtime/tests/runtime_integration_tests.rs
bajtos: > Why do we not unwrap an error once it occurs?

In some tests, we want to assert that execution of the JS code threw an error and that the JS code produced the expected logs.

I find it easier when all test helpers use error propagation instead of unwrapping, because that way I don't have to decide in each test helper which error-handling strategy is appropriate.

Let's continue this discussion in https://github.com/CheckerNetwork/zinnia/pull/752

PR #741, opened 2025-04-17
chore: update list of ignored security advisories

bajtos:
- Stop ignoring advisories that are no longer relevant.
- Ignore new advisories that are in dependencies outside of our control.


bajtos:
👌🏻 

Screenshot 2025-04-17 at 16 14 12

PR #747, opened 2025-04-23
docs: verify Linux release artifacts run in Docker

bajtos:
Capturing the comment https://github.com/CheckerNetwork/zinnia/pull/172#issuecomment-1506641021

### DIFF
--- a/docs/release.md
+++ b/docs/release.md
@@ -35,3 +35,42 @@ refer to `X.Y.Z` as _major_, _minor_ and _patch_ version.
    changes.
 
 1. Click on the green button `Publish release`
+
+## Testing the release build pipeline
+
+### Linux binaries
+
+We need to link the Linux binaries with a glibc version that's compatible with the target system, e.g. the Node.js docker image used by Checker Node.
+
+Verification steps (assuming Apple Silicon/arm64 machine):
+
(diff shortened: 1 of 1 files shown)


docs/release.md
```diff
@@ -35,3 +35,42 @@ refer to `X.Y.Z` as _major_, _minor_ and _patch_ version.
    changes.
 
 1. Click on the green button `Publish release`
+
+## Testing the release build pipeline
```
juliangruber, line 39: Is this already part of CI? Or is this something we're expected to perform manually before publishing?

bajtos: Great question!

This is not part of CI, we're expected to perform this manually when making changes to the release build pipeline (the GitHub Actions workflow).

I.e. this is a manual check to verify that the release build pipeline works as expected.

Do you think it's worth automating this check and running it for every release?

juliangruber: I think it's fine to keep it manual, but can you please document that this is to be performed when working on the pipeline? That wasn't clear to me

juliangruber: Ie "testing the release build pipeline" could also mean it's something you're supposed to do after every release, to verify the release build pipeline worked

PR #752, opened 2025-04-24
refactor: run_js_test_file in integration tests

bajtos:
Rework the helper `run_js_test_file()` to return `Result<Vec<String>, RunFailure>`, which makes it more ergonomic to handle run errors in tests expecting the run to succeed.

This is a follow-up for the discussion we had in https://github.com/CheckerNetwork/zinnia/pull/740#discussion_r2049002936


pyropy (approved):
Nice refactor, code looks much cleaner! 👍🏻

